PR #21, opened 2015-02-01
Add Saitek P2600 config udev.

xoddark:
Tested with Lakka on Cubox-I

Signed-off-by: XoD <xoddark@gmail.com>


twinaphex:
Hi, thanks for this.

Might be an idea to add VID/PID in a later commit to it too.

xoddark:
Ok, how I can know the VIP/PID ?

Kivutar:
Do a lsusb, you will see two hexadecimal numbers on the line of your joypad. The first one is vid, the second one is pid. You have to convert them to decimal and put them in the config file.

PR #25, opened 2015-02-21
Add udev autoconfigs from retropie

gizmo98:


Kivutar:
That's great. We could convince them to use this package maybe?

Kivutar:
And isn't the PS3Controller a duplicate?

(also, they all lack vid:pid)

gizmo98:
Yeah. I added this repo lately and it should be easier to maintain one repo for all.

The new PS3Controller config is necessary if you pair controller with bluez 5.xx with sixaxis support. The old configs work if QTSixA is used.

gizmo98:
Is vid:pid necessary now? I don't have all of these controllers. It's a collection of files we have collected over the years.

Kivutar:
vid:pid are not exactly needed, but when you have two controllers that declare the same name, it's the only way to differenciate them. It happens sometimes with chinese clones.

iirc, in retroarch vid:pid matching takes precedence over name matching.
@twinaphex can you confirm?

Anyway it's a very good thing that we share this repo.

PR #32, opened 2015-03-11
Add partial support for raphnet.net 4nes4snes

dnrce:
http://www.raphnet.net/electronique/4nes4snes/index_en.php

This is only the first 2 axes and 4 buttons.

This is a multi-controller adapter. Is there a good way to add support for the other three ports? It presents as a single 8-axis, 32-button controller.


twinaphex:
> This is a multi-controller adapter. Is there a good way to add support for the other three ports? It presents as a single 8-axis, 32-button controller.

Not sure to be honest, I don't think we've encountered such usecases yet.

Is there any differentation you could do based on VID/PID in combination with the device name? Does the device name change when attaching a different pad to it? You'd have to do some experimenting here on your end.

dnrce:
@twinaphex, no, it's a single device.

On a related note, should I add the USB vendor/product IDs here?

twinaphex:
Definitely add the USB vendor/product IDs to the autoconfig file. It provides for a more advanced way of detection vs. a simple name match that can lead to false positives.

PR #42, opened 2015-04-19
Create USBgamepad.cfg

HerbFargus:
Configs for a cheap SNES USB controller


twinaphex:
Can you provide VID and PID values for this too? This is to prevent false positives since the input device name is fairly generic.

HerbFargus:
Updated

twinaphex:
Can you change those values to decimal?

Alcaro:
What is o81f in decimal?

twinaphex:
Still doesn't have decimal values for VID/PID, but whatever I guess.

HerbFargus:
just changed it using https://www.branah.com/ascii-converter and got decimal values- is that what you are looking for? I just committed the changes, right after you merged it, should I submit another pull request?

twinaphex:
Yes, submit another PR.

HerbFargus:
done

PR #43, opened 2015-04-19
Update USBgamepad.cfg

HerbFargus:


Alcaro:
I don't know what exactly you did to those numbers, but the correct decimal values for those two are 2079 and 58369.

HerbFargus:
I've changed it to your recommendations. sorry, I'm not super versed in ascii conversion, I'm new to all of this. where are some resources that you would recommend I learn how to do that?

Alcaro:
A good first step would be to discard all thoughts about ASCII. This is hexadecimal (base 16) and decimal (base 10), nothing else.

twinaphex:
If you have Ruby installed, you could just type irb. You could then input a hexadecimal value and it presents you with the value in decimal. 

Alternatively, go to an online calculator that can convert between hexadecimal and decimal or do it by hand. Anyway, I'll merge this.

HerbFargus:
ok. Thanks. sorry about that

PR #46, opened 2015-05-06
Create GS_gamepad.cfg

gamerminstrel:
Gamestop Red Samurai Gamepad support for when the controller is in gamepad mode.


gamerminstrel:
It took some poking around, but I managed to make this work. Did I do this right?

PR #51, opened 2015-05-26
Add Mayflash-GC-to-USB-Adapter_Nintendo-GameCube-Controller.cfg

paalfe:
Genuine Nintendo GameCube Controller connected to a Mayflash GC to USB Adapter (2 ports).

I Did expereince problems because of HuiJia__USB_GamePad.cfg that has the same input_device name. They have different VID:PID, only the input_device name is the same. Had to remove HuiJia__USB_GamePad.cfg from my autoconfig directory.


hizzlekizzle:
So this doesn't work unless you remove the HuiJia pad cfg?

Kivutar:
It means that retroarch detection code have to be modified to give more weight if both the name and the vid:pid matches.

The current behavior is that it will stop to loop over the config files when one with either name or vid:pid is found.

paalfe:
I see Mayflash has several (older and newer) USB adapters that identify them with the same input_device name (HuiJia USB GamePad), but they do have different vid:pid. So this problem will probably happen again and RetroArch detection code have to be modified to avoid this. I have found other people struggling with this same issue, here is a forum thread in the libretro forums of someone experiencing this problem http://libretro.com/forums/showthread.php?t=1511.

twinaphex:
@kivutar are you going to merge this?

Kivutar:
Now that we have a weighted system, it should fix all these issues alltogether. Don't you think @paalfe ?

Even if Mayflash adapters share the same name, they have different vid:pid and RA is now able to detect that.

With the latest RA, if you keep the HuiJia config in your directory, do you experience problems?

paalfe:
Running v1.2.2 and it works, it can be added :+1:

PR #69, opened 2015-07-11
Update and rename 8BitdoNES30GamePad.cfg to 8Bitdo-NES30-GamePad-Blue…

paalfe:
I was wrong in my previous commit (bb39fd7acca30b9075c0437b2b556351e30eb6f7) about the button assigning being wrong, in jstest-gtk they show as the button layout from my first commit, but in RetroArch they show as they are assigned in the original cfg.

In my last commit (783ad964ff85f62aa79569b978e009b0eb0301f9) I have reverted to the original cfg button assignment, added vid:pid and labels.

I recommend renaming the cfg since I have made a pull for the same controller connected with USB and I have added more 8Bitdo controller configs, would be nice if they where easier to find. More 8Bitdo controller configs to come... 8Bitdo has great controllers :D


paalfe:
I have done proper testing and autoconfig with this cfg works, pull request should be accepted.

hizzlekizzle:
Radius just committed a bunch of stuff for the 8bitdo controllers. We should have him take a look at it first, but it seems fine to me at a glance.

paalfe:
It was me who committed all those 8Bitdo controller configs :D

fr500:
I commited Windows profiles

paalfe:
It was twinaphex who committed all my other udev 8Bitdo controller config pull requests.

hizzlekizzle:
k, sounds good.

PR #112, opened 2015-09-03
Create 8Bitdo NES30 GamePad.cfg

CaptainAlphabet:


hizzlekizzle:
does this conflict with:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/android/8Bitdo%20NES30%20GamePad%20Joystick.cfg
?

fr500:
Well it seems VID/PID are the same but this one was made with it running in the default mode (which is unsuitable for gaming in some platforms since it just emulates keyboard arrows)

That said, even thought the IDs are the same the names are different, so with the current score based system it should work fine with both in 5.0+ devices

hizzlekizzle:
kk, merging it, then. Thanks for the input :)

CaptainAlphabet:
Thank you.

A possible solution for conflicting or wrong selected autoconfig files could be:

[...] a "User 1 Load Autoconfig" feature beneath the "User 1 Save Autoconfig" which brings up the autoconfig directory for selecting and applying a specifig autoconfig file.

http://libretro.com/forums/showthread.php?t=3550&page=3&p=26659&viewfull=1#post26659

PR #118, opened 2015-10-05
Updated all 8Bitdo configs. Added info/comments and rearranged the layout of the configs. Similar to the 8Bitdo Zero config.

paalfe:
Updated all 8Bitdo configs. Added info/comments and rearranged the layout of the configs. Similar to the 8Bitdo Zero config.


KuriGohan-Kamehameha:
how do I save these files and where do I put them? lol signed up just to ask this. trying to get NES30Pro working with retroarch android

fr500:
The NES30 PRo on the latest firmware works out of the box

KuriGohan-Kamehameha:
OK, I updated recently but I'll see if there's a newer version out there. Thanks! Ps, I'm trying to get the thumb sticks working for n64

--- Original Message ---

KuriGohan-Kamehameha:
Ok I'm on 1.66 firmware but only the dpad works. I can't press A or B or use the thumb sticks.

fr500:
on what device?

KuriGohan-Kamehameha:
2013 nexus 7

--- Original Message ---

KuriGohan-Kamehameha:
I am using the mode where you just press POWER and the blue light blinks 1/s

KuriGohan-Kamehameha:
N64oid recognizes all of the buttons except the right thumbstick. It even recognizes the button on the thumb stick, but not the thing itself.

fr500:
N64oid does things differently that doesn't concern RetroArch at all.
I don't know what's wrong on your setup but you can try binding the buttons manually like explained here: http://www.libretro.com/index.php/getting-started-with-retroarch/

If it's recognizing the pad it should print an 8bitdo nes 30 pro configured message at startup.
You haven't mentioned what retroarch version are you using.

KuriGohan-Kamehameha:
1.0.0.2, I think. It may be referring to the version of the core

fr500:
1.0.0.2 is ancient, we don't support that anymore

KuriGohan-Kamehameha:
Ok, I don't use the play store, so maybe that's why I got an old version. Do you have a link for the apk of the newest version?

KuriGohan-Kamehameha:
Ok, I found 1.2.2, and the controls seem to be working, but now the frames are killing me! It's like running at half speed as if my device can't handle the game, even though the old retroarch and n64oid worked perfectly fine in that regard

fr500:
It is explained on the guide I linked

KuriGohan-Kamehameha:
I don't see it mentioned, maybe I'm just stupid. Sorry!

KuriGohan-Kamehameha:
I tried something with threaded video and running a setting for 2048 frames and it didn't help. Still very laggy

KuriGohan-Kamehameha:
I set the GFX plugin to rice and it didn't help

KuriGohan-Kamehameha:
The controls are definitely working, but the speed is laggy even with rice. I kept on alternating between n64oid and retroarch to make sure I'm not the problem. Retroarch has fully functional controls but is laggy and n64oid runs perfectly but the right thumb stick doesn't work.

twinaphex:
N64oid is an ancient version of mupen64plus with an even more ancient and hacked up graphics plugin.

paalfe:
These cfg files is for udev (linux). You find Android cfg files at https://github.com/libretro/retroarch-joypad-autoconfig/tree/master/android.

paalfe:
If RetroArch (on linux) do NOT detect a controller when connected via Bluetooth and won't even recognize manual binding, fix this by adding these udev rules to force detection - https://github.com/paalfe/mixedcontent/blob/master/udev_rules.d/99-8bitdo-bluetooth-controllers.rules. 
Add the udev rules file to "/etc/udev/rules.d/99-8bitdo-bluetooth-controllers.rules" and reboot. 
Forum thread: http://libretro.com/forums/showthread.php?t=3649

paalfe:
If installing RetroArch on Ubuntu, these cfg files is included in the "retroarch-joypad-autoconfig" package from https://launchpad.net/~libretro/+archive/ubuntu/stable 
You find the retroarch-joypad-autoconfig path in the RetroArch settings menu or in the RetroArch config file, "\usr\share\libretro\autoconfig\" is the default retroarch-joypad-autoconfig path.

KuriGohan-Kamehameha:
Ok I get that n64oid is ancient and outdated, but right now it's working better than Retroarch because it's not lagging. Retroarch Android v1.0.0.2 worked better than 1.2.2 does, even though 1.2.2 has full controller support, which is what I came here looking for. Does anyone have a solution?

twinaphex:
As far as i understand it, Mupen64plus mainline has just become slower due to recent refactors.

I cannot help it if upstream projects make themselves slower due to refactors and don't care about the consequences. As a matter of fact, I got plenty of grief whenever I did care and forked projects altogether, then I got accused by these projects of not wanting to stay inline with upstream etc., so there is no pleasing people. But as usual we get blamed regardless by endusers who don't understand how things work.

Bottom line, I don't think I'm going to resort to old crappy code just because your hardware is totally underperforming and can't handle mupen64plus mainline. If I can find optimizations that work out well on your hardware, I am always there to merge it. But I'm not going to go out of my way to specifically cater to whatever outdated hardware you seem to own, sorry.

KuriGohan-Kamehameha:
C'mon, now! It's not like I was complaining, I'm just trying to find a solution. If that's the reason, then so be it, I'll either make it work in some back-alley way or I'll find an alternative. Just so you know, a 2013 Nexus 7 isn't outdated, so I don't think that's the issue. I agree, sometimes apps get more bloated and/or need more power to do the same thing when you start adding small features here and there. I'll stick to whatever works the best for my device then, probably N64oid, since it is functional in both input and video, albeit not perfect. Thanks for all your help, guys!

twinaphex:
I told you before, this is not the right place to be posting about issues you are having with Mupen64 libretro. There is a separate github repo for that.

twinaphex:
There, removed all the noise that was pertinent to the subject matter at hand. Sorry to everyone involved in this thread.

PR #120, opened 2015-10-06
Add Xiaomi Gamepad config

EugenGanshorn:


twinaphex:
This is causing the Android build to no longer build on our buildbot. Please remove these Chinese characters out of the filename and send a new pull request again.

PR #128, opened 2015-11-14
udev autoconfig for HORI Fighting Commander 4

duganchen:
I've tested this with Street Fighter II: Championship Edition and with Golden Axe, both using the Genesis Plus GX core.


twinaphex:
Hi, thanks for this, I'll merge it.

Can you also add button labels to this later so that it shows them in the menu?

PR #130, opened 2015-11-14
osx : fix hid xbox 360 profile according to pull request #2391 inside retroarch project

Arche-san:
Hello,

I made a fix inside retroarch project to manage right stick for xbox 360 in os x.
Some profiles need to be updated for hid drivers. Here's a fix for xbox 360 gamepad.

Cheers ! :)


twinaphex:
Hmm the thing that concerns me here is:

the hid joypad autoconfig profiles are not just used on OSX, but also on Linux (through libusb), and also on the Wii. on Mac IOHIDManager gets used instead.

By changing these values, it would cause the HID drivers for Linux and Wii to no longer work unless they were changed as well. I think it would be better if we could prevent modifications to the auto config files and try to solve it another way (possibly at the RetroArch level instead). The HID autoconfig profiles should remain platform-agnostic.

Arche-san:
Understood ! :)
If you want I can change axis order values for os x into retroarch project to keep compatibility everywhere. But i need to know the right order.
Here's an example : 
+0/-0 : Left Stick Horizontal
+1/-1 : Left Stick Vertical
+2/-2 : Right Stick Horizontal
+3/-3 : Right Stick Vertical
+4/-4 : Left Trigger
+5/-5 : Right Trigger

I think trigger buttons must be the last values for config files because all gamepad drivers don't manage triggers with axis (it must be specific to xbox360 gamepad and maybe ps3/ps4 gamepad).

If you know the right order for linux/wii config please let me know and I'll create a pull request with the valid order into retroarch project for os x. I also need to fix the xbox 360 profile : i've created it 20 days ago and i'm pretty sure the initial values are not correct for the right stick (not detected until my fix).

twinaphex:
Left Stick horizontal/vertical, Right stick horizontal/vertical check out at least.

As for left and right trigger, well, depending on the pad, it can either be assigned to axis values (like +4/+5) or just plain digital buttons. The RetroPad that they get mapped against would just treat the trigger buttons as regular digital values anyway.

Anyway, you can look at the current autoconfig files inside the 'hid' folder. These autoconfig files are used by OSX/Linux/Wii. So those exact values I suppose. 

If you need help identifying the exact files involved with each platform, let me know. You can also reach me directly on IRC, I hang out on #retroarch (Freenode) as Twinaphex, you could PM me there and I could help answer questions if you have any.

Arche-san:
Alright!
I made a fix into retroarch project to keep compatibility everywhere. If everything is ok you don't need to edit all others profiles! :)

I'm going to close this pull request and create another one to fix xbox 360 hid profile only.

PR #133, opened 2015-12-08
Add some autoconfig profile for dinput, xinput, and sdl2

Monroe88:


fr500:
Thanks, the xinput are identical to stock builtin binds right?

Monroe88:
Yes they're identical, except they have button descriptors added to them.

PR #134, opened 2015-12-28
Updating Logitech_Gamepad_F310.cfg

alexbain:
I found that my Logitech Gamepad config differed from the one stored in the repository. This configuration file now matches what was generated when I ran through the UI configuration.


Kivutar:
Sadly, there are a ton of Logitech_Gamepad_F310 in the market with different settings :(

nschloe:
I have the same controller, and my setup is indeed different. For example
```
input_l_btn = "4"
```
is still present here after configuration.

Kivutar:
So, what do we do? We store each different configuration here and let the users remove the one not corresponding to theirs? We ban all F310 autoconfigs on this repo? We choose the most common one?

fr500:
are the VID/PID the same in all cases?
also, why does the autoconf have state slot increase/decrease? we only want retropad mappings here

fr500:
It seems this path is xinput compliant, if that's the case, maybe it has a dinput/xinput switch and the configs were made with the switch on a different setting?

udev/Logitech_Gamepad_F310.cfg
```diff
@@ -2,20 +2,27 @@ input_device = "Logitech Gamepad F310"
...
 input_left_btn = "h0left"
+input_state_slot_decrease_btn = "h0left"
 input_right_btn = "h0right"
-input_a_btn = "1"
-input_x_btn = "3"
-input_l_btn = "4"
-input_r_btn = "5"
-input_l2_axis = "+2"
-input_r2_axis = "+5"
+input_state_slot_increase_btn = "h0right"
```
fr500: This is no good, this is too user specific. We only want retropad binds here, not hotkeys.
Also this map doesn't have descriptors.

fr500:
For reference on descriptors use this:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/XInput%20Controller%20%28User%201%29.cfg

The second section has the actual button names, that's what we want in all the contributed profiles

alexbain:
Closing this since it looks like the existing config is more generic.

PR #138, opened 2016-01-05
Add direction inputs for D-Pad as buttons

RobLoach:
Also rename to Steam_Controller_udev.cfg so that autoconfig building is not overwritten by the `linuxraw` configuration.


hizzlekizzle:
I'm not sure I follow your reasoning on the rename. Each driver's configs go into their own directory, so this shouldn't get overwritten, even if they have exactly the same name.

PR #149, opened 2016-01-22
rename 8Bitdo FC30-1P and FC30-2P configs

paalfe:
rename so all FC30 configs is listet after each other, together with 8Bitdo-FC30-GamePad-USB.cfg


paalfe:
merge?

PR #193, opened 2016-05-22
Two new controller auto configs

drinimar:
Added two new controlers, wireless PS3 from 5Star and wired PS3 from Microntek (with spaces, not underscores).


hizzlekizzle:
Those look good, and I'll merge them, but while we have you, would mind adding button labels, as well? Here's an example:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/MP-8800_Quad_USB_Joypad.cfg

drinimar:
Should be good now.

hizzlekizzle:
Beautiful! Thanks for adding those :)

PR #194, opened 2016-05-28
Nvidia shield controller config

R3d-Sky:
NOTE: This file requires an udev rule AND the bottom touchpad is mapped to select due to the home and back buttons being situated on an emulated keyboard (software-wise).


fr500:
Please look for recent commits and add the descriptors like on those. 
We are not merging profiles without input descriptors

R3d-Sky:
I've just added the input descriptors. Can PR be merged now?

PR #195, opened 2016-05-31
Update 8Bitdo_ZERO_BT.cfg

paalfe:
I have two zero gamepads, one blue colored (bought when it was introduced) and one famicom edition (red colored). They both is identified as "8Bitdo Zero GamePad" with Decimal vid:pid = 1440:12850, but they have different mappings. There is no firmware release available and I'm not sure if there is different mappings of the blue colored gamepad or if the difference is only blue colored vs. famicom edition.


paalfe:
I think this should be merged because most people will probably have zero gamepad requiring this mapping cfg.

hizzlekizzle:
i have a newer zero, so I can double-check the new mapping.

Kivutar:
Ok then

PR #197, opened 2016-06-13
Add Retronic Design USB joystick adapter for Sega Genesis/Megadrive

gouchi:


gouchi:
udev configuration file.

PR #201, opened 2016-06-19
Add udev config files

orbea:
Adds a config file for my gamestop brand usb xbox360 controller.


orbea:
I added a commit for my Smartjoy snes to usb converter.

hizzlekizzle:
Those look good! Would you mind adding the button labels, as well? This is a good example of how that looks:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/android/Amazon%20Fire%20Game%20Controller.cfg

orbea:
Sure, I added two more and will add button labels now.

orbea:
I added the button labels as well as removed some unused buttons found on some of the controllers. The only thing that may be missing is the home buttons on some controllers which are not added when the autoconfig is generated? And also the turbo button on the Hori fightstick which does not seem to work at all...

hizzlekizzle:
Good work! Thanks for these :)

PR #202, opened 2016-06-19
Create Sminiker.Gamepad.cfg

oddy999:


hizzlekizzle:
Looks good! While we have you, would you mind adding the button labels, as well? This is a good example of how it looks:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/android/Amazon%20Fire%20Game%20Controller.cfg

hizzlekizzle:
Perfect! Thanks :)

PR #204, opened 2016-06-22
Add config for DS4 connected through manual mode to Bluez 5

parkerlreed:


hizzlekizzle:
Looks good, but would you mind also adding in the button labels? This is a good example of what they look like:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/Sony-PlayStation4-DualShock4-Controller.cfg

parkerlreed:
Ahh can do.

parkerlreed:
Done

hizzlekizzle:
Looks great. Thanks!

Kivutar:
@parkerlreed I also have a late DS4 controller, and it matches your config file when I connect it with bluez. However, something seams to be blocking on the kernel or bluez side : when I cat /dev/input/js0 I only get signals for the trackpad, not for the buttons. Any idea?

https://github.com/libretro/Lakka/issues/403

parkerlreed:
Huh, yeah mine works fine. 

Model CUH-ZCT1U

Are you running a recent Bluez? 5.14 is what added full support.

````
Unless your system is using BlueZ 5.14 (which was released recently) or higher it is not possible to pair with the DS4. Therefore this workaround exists, which connects directly to the DS4 when it has been started in pairing mode (by holding Share + the PS button until the LED starts blinking rapidly).
````

From https://github.com/chrippa/ds4drv

Other than that try pairing the same way and then running `ds4drv --hidraw` Should have the same net effect of making it show up as a regular DS4 controller.

Kivutar:
I'm running bluez 5.40 here on my arch linux.

Mine is CUH-ZCT1E

Do you know if your controller works on Lakka?

parkerlreed:
Have not tested. I am on the same Bluez version in Arch... I have the C1 at work, how do I go about trying it in Lakka?

parkerlreed:
https://www.reddit.com/r/PS4/comments/3g2h7z/question_what_is_the_difference_between_cuhzct1e/

That is suggesting I have the US model and yours is EU. Wonder if there's some hardware difference.

PR #206, opened 2016-06-25
Create HuiJia  USB GamePad.cfg

markwkidd:
[HuiJia  USB GamePad.zip](https://github.com/libretro/retroarch-joypad-autoconfig/files/332846/HuiJia.USB.GamePad.zip)

This autoconfig matches a 2-port SNES to USB adapter sold under the Mayflash brand name.


hizzlekizzle:
That looks good, but while we have you, would you mind adding the button labels, as well?

Here's an example of how that looks:

markwkidd:
No problem. Please feel free to check my update to see that I have the format correct.

And while I have your attention :)

... I'm helping update the Lakka documentation. I don't think that what we have now in terms of how contribute joypad autoconfigs is going to get many users all the way to the point of doing a pull request, which is what the lead developer would like.

Here's the current doc: https://github.com/libretro/Lakka/wiki/Contributing-your-joypad-config

At the moment I have planned to add a few screenshots showing how to add a new file to the udev folder and create a pull request.

**My question for you**: do you know of any other existing docs or good forum posts out there that describe the process better than what we have now? Especially anything like creating button labels that should be done by hand after RetroArch spits out the config file.

I'm not particularly familiar with RetroArch -- I'm just a Lakka user who's a technical writer in a different domain. I'm hoping someone else has already described what needs to be added in terms of button labels, etc. somewhere else and I can use that as a first draft.

Thank you! And do let me know if I need to change anything else with the config.

hizzlekizzle:
There's a sticky in the Android subforum that talks about making autoconfig files but I don't know of any other documentation, no. I don't actually know when the button labels were added or if there was any fanfare or documentation that came along with it, I just started noticing them in some of the files and decided to start asking for them on PRs I merge.

PR #209, opened 2016-06-30
USB controller for PC

lollo78:


hizzlekizzle:
While we have you handy, could you add in the button labels, as well? It makes remapping much easier for users. Here's an example of what they look like:

lollo78:
of course

PR #210, opened 2016-07-01
USB Controller for PC with button label

lollo78:
I added button label like required by hizzlekizzle: 
https://github.com/libretro/retroarch-joypad-autoconfig/pull/209


hizzlekizzle:
Looks good. Thanks!

FAKEDIT: looks like this is identical to an earlier-submitted pad but with more buttons. It should be fine to delete the old one, so I'll merge this one and delete the other.

lollo78:
yes the old one was not complete form me. So I had to make this

PR #215, opened 2016-07-19
Create Retrolink_(S)NES_USB.cfg

moTechPlz:
Config file for both the retrolink NES and retrolink SNES controller. The controllers both use the same product id as retrolink is known to do. However the NES controller maps it keys to the same scancode as the SNES controller does, so this config works for these two controllers, Other products or future products by retrolink using the same product id may be a problem.


hizzlekizzle:
Looks great! Thanks :)

PR #217, opened 2016-08-06
Added configuration for Mad Catz FightStick TE2+ in both PS3 and PS4 mode.

Hiki1:
FightStick works both when switch is in PS3 and PS4 mode.
For directions input only DP switch is now supported.


hizzlekizzle:
looks good! thanks for including the labels! :)

PR #222, opened 2016-08-16
Added config for Razer Serval

hooby3dfx:


hizzlekizzle:
Looks good, but while we have you, would you mind adding button labels, as well? This one shows what it looks like:

hooby3dfx:
Done - I need to figure out the menu button mapping then it should be good.

hizzlekizzle:
Those labels look good. I'll go ahead and merge this and you can send another PR when you get the menu figured out.

PR #247, opened 2016-10-30
Added generic NES and SNES like USB controllers

magwe:
These controllers share the same USB ID but have different button
configurations.

### DIFF
--- a/udev/usb_gamepad___________(NES).cfg
+++ b/udev/usb_gamepad___________(NES).cfg
@@ -1,0 +1,35 @@
+# 0810:e501 Personal Communication Systems, Inc.
+
+# This is a NES style gamepad that uses a generic chinese USB gamepad encoder
+# chip common in other gamepads with different button configurations.
+
+input_driver = "udev"
+input_device = "usb gamepad           "
+input_vendor_id = "2064"
+input_product_id = "58625"
+
+#########
+
+input_a_btn = "1"
+input_b_btn = "0"
+input_select_btn = "8"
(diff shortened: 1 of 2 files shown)


Kivutar:
If they share the same vid:pid and device names, but don't share the same button config, I don't see how autoconfig can select them properly.

magwe:
You are right, autoconfig will just grab one of them. You would have to rename unwanted configurations manually to make sure to get the right one. Bot now you have the option to do that atleast.

This is a general problem. The other pull request I made that was merged just before this one has the same issue. It is for "0079:0006 DragonRise Inc.   Generic   USB  Joystick" which is another of these common generic chinese usb encoders used by many different cheap game controllers.

Most of them are "dual shock" style, but for example the "Retrolink N64 USB controller" share the same VID:PID. I don't own that controller, but the .cfg says that the name is "Retrolink Nintento 64 Classic USB". I can't verify that but it sounds strange to me that the name is different but the PID is not. (Is the name stored on the controller or does it come from software?)

My Dragonrise controller (an arcade stick I build myself with the Bosega controller board) sometimes identifies as "  Generic   USB  Joystick" and sometimes as  "PC TWIN SHOCK Gamepad" depending on application. This is yet another controller with the same VID:PID. the "PC TWIN SHOCK Gamepad" has a separate .cfg but the layout is identical to "  Generic   USB  Joystick".


One of these

Kivutar:
As long as it doesn't break autoconfiguration for a wide spreaded controller, it's ok I guess.

PR #251, opened 2016-11-03
xinput 8Bitdo configs: update and confirm working with latest firmware.

paalfe:
Includes 8Bitdo N64 USB and BT configs.


paalfe:
Ready for merging.

PR #252, opened 2016-11-04
Create ibuffalo_classic_usb_gamepad_alternate.cfg

minipretzel:
Note the spaces in input_device.. unsure if original iBuffalo_Classic_Usb_Gamepad.cfg has typo ie no spaces.  Different vendor/product ids.


hizzlekizzle:
if the vid/pid is different, I'm all for it.

PR #259, opened 2016-11-20
[xinput] Don't swap menu ok and cancel buttons on controllers with Nintendo layout

tony971:


fr500:
Did you even test this? It's not implemented you can't just load random stuff when loading autoconfs

tony971:
I figured adding this to a config file isn't harmful. Have it in wait until it actually does something.

PR #260, opened 2016-11-22
Added support for the Playstation 3 controller.

mudlord:
Added support for a generic PS3 controller clone.


hizzlekizzle:
Looks good! Would you mind adding button labels while you're at it? Here's an example:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/PLAYSTATION(R)3Conteroller-PANHAI.cfg

PR #263, opened 2016-12-03
Add support for the RetroUSB NES, SNES, and Genesis adapters

dvsayson:


xinput/RetroUSB_RetroPort_SNES.cfg
```diff
@@ -0,0 +1,33 @@
...
+input_y_btn_label = "Y"
+input_select_btn_label = "Select"
+input_start_btn_label = "Start"
+input_up_btn_label = "D-Pad Up"
+input_down_btn_label = "D-Pad Down"
+input_left_btn_label = "D-Pad Left"
+input_right_btn_label = "D-Pad Right"
+input_a_btn_label = "A"
+input_x_btn_label = "X"
+input_l_btn_label = "Left Shoulder"
```
fr500: these should be just L and R in SNES, other than that it seems ok

PR #271, opened 2016-12-27
Add ChromeOS Gamepad for Android on ChromeOS

denniskempin:
All gamepads on ChromeOS are mapped by Chrome and passed to a virtual
device in Android. The virtual device is always named "ChromeOS Gamepad"
and follows the standard Android mapping.


denniskempin:
@twinaphex Do you match gamepads by name if pid/vid are not available? Since devices on chromeos are virtualized, they do not have a product/vendor id.

twinaphex:
There's a name match attempt yes.

Can I take it you verified that this autoconfig works?

PR #273, opened 2016-12-28
Adds DS4 autoconfig profile with standard Windows 10 drivers.

Schroedingers-Cat:
+Tested on Windows 10, x64
+Works via USB and bluetooth connection


Kivutar:
thanks

PR #275, opened 2016-12-29
Add MAYFLASH Arcade Fightstick F300

neersighted:


hizzlekizzle:
These look good, thanks! While I have you, would you mind adding the button labels, as well? That makes it easier to tell which buttons are which during core input remapping.

neersighted:
The labels on the Fightstick correspond directly to the default button names, so unless there is some benefit to the full label (X is labeled 'X/Square' for instance), I don't see any real benefit from manually specifying the button names.

hizzlekizzle:
Ok, that works :)

Thanks again!

PR #298, opened 2017-01-09
Rename Logitech_Logitech_Dual_Action.cfg to Logitech_Dual_Action.cfg

paalfe:


hizzlekizzle:
Thanks for all of these, by the way!

PR #320, opened 2017-01-23
Adds support for Mayflash WiiU Pro Game Controller Adapter

daaveuk:
I've created an autoconfig files for the dinput mode for this adapter.

http://www.mayflash.com/Products/NINTENDOWiiU/W009.html

The advantage over the Xinput mode is allows up to four Wiiu Pro controllers to connect and LAKKA seems to detect each controller automatically with this configuration in place. It even seems to assign the player status light to the correct controller. I've managed to test it with both a genuine controller and a 'off-brand' imitation and both work perfectly.


hizzlekizzle:
Ah, that sounds nice! Would you mind adding the button labels, as well, while we have you?

daaveuk:
Happy to :) I'll have a look through and see what I can do. I'll also try and get the menu button mapped to the Lakka menu while I'm there!

twinaphex:
Yeah, you can make another PR adding those labels.

PR #345, opened 2017-05-24
Add autoconfig for 'Speedlink SL-6603-TWT Competition Pro USB Joystick'

ltrvs:
- Similar as: http://www.speedlink.com/?p=2&cat=414&pid=25585&paus=1
- Label in the joystick states: "Speedlink SL-6603-TWT Competition Pro USB Joystick"
- The .cfg file is unmodified, generated with Lakka
- Spaces in the filename replaced with underscores to match the existing "MOSIC______USB_2A8K__GamePad__________.cfg" filename pattern


hizzlekizzle:
That looks good so far. The only other thing I would request is button labels, for example:
input_b_btn_label = "Cross"
input_y_btn_label = "Square"
That will make the 'controls' menu display the actual name of the button when people use input remapping instead of just showing the assigned retropad button.

ltrvs:
Thanks for the feedback. I'll give a try for the labels, probably in a few days.

Since the device's buttons are blank and doesn't follow the ABXY gamepad layout (please see the link above for details), are there any suggestions for the labels?

Im thinking something like "Left round"/"Left triangle" OR "Big left"/"Small left". I'm not a native English speaker, so please let me know if these sound silly :-)

hizzlekizzle:
Hmm, yeah, that's tough... I guess, yeah, something like round 1, triangle 2 or Left/Right Round/Triangle would be most descriptive.

ltrvs:
Labels added.

(I can see them in the menu when loading default binds, but when assigning custom ones the menu shows plain numbers. Assuming this behaviour is as intended.)

hizzlekizzle:
Yeah, looks good to me! Thanks for the additional effort :)

PR #346, opened 2017-06-03
Add files via upload

Jusix:
This device also gets used in N64 controller which have a different mapping


Kivutar:
We already have 3 occurences of this vid:pid:name in the udev folder.

PR #348, opened 2017-06-09
Fixes faulty DS4 wired linux profile and adds bluetooth linux profile

Schroedingers-Cat:
On Ubuntu 17.04 using latest stable retroarch 1.6.0, the DS4 doesn't work, neither in wired mode nor via bluetooth. Retroarch simply does not react to its input. Using the DS4 via USB even blocks the keyboard input when retroarch ist started after the controller has been connected. Furthermore, when connecting the controller via USB while retroarch is running, the inputs of both the keyboard and the controller are acting mad (pressing a key arrow or the d-pad once skips a lot of entries).

This MR fixes these issues by using the correct axis values (udev only) and adds a profile when connecting the DS4 via standard bluetooth utility.


Kivutar:
Did ubuntu patch their kernel or something? The DS4 is working fine on both mode on Arch Linux (even on older kernels like 4.8.8).

I'm going to test this change, but I think the vid:pid included your bluetooth version is the vid:pid of your bluetooth dongle, not the one of the pad. Can you correct this please?

Schroedingers-Cat:
Ubuntu 17.04 uses a more recent kernel. Maybe that explains the difference?
```bash
uname -srvo
Linux 4.10.0-22-generic #24-Ubuntu SMP Mon May 22 17:43:20 UTC 2017 GNU/Linux
```

How can I find out the vid:pid? lsusb only shows usb devices, so the bluetooth connected ds4 doesn't show up there. Also, the reported vid:pid strings of my bluetooth dongle are `0a5c:21e6` which does not match the uploaded configuration file entries.

Kivutar:
You can just copy the vid:pid from the USB version. This will avoid a bug where people that are using the same USB dongle as yours get this config selected for any pad.

Schroedingers-Cat:
Ok, this should fix it.

Kivutar:
Now it's me who is getting this non working DS4 + blocked keyboard when Sony-PlayStation4-DualShock4-Controller-Bluetooth.cfg is selected.

twinaphex:
This makes my entire system hard freeze now when I unplug the PS4 joypad.

Reverting this. It really sucks that they screwed up DS4 input ever since Linux 4.10. Very bad regression.

twinaphex:
Correction - this might have not been the problem. I can probably revert it later once I find the real cause of all these problems.

twinaphex:
My fault.

I indeed accidentally caused a regression. I have now fixed that issue.

Going to re-merge that PR.

Schroedingers-Cat:
@Kivutar so it sounds like retroarch either needs to support kernel-based button mappings or someone should investigate why DS4 button mapping had to change (maybe it was unintentional).

Schroedingers-Cat:
It could have happened here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/hid/hid-sony.c?id=9131f8cc2b4eaf7c08d402243429e0bfba9aa0d6

Kivutar:
I wonder why it breaks keyboard. Even after upgrading RA master with the bugfix of Twinaphex.

Schroedingers-Cat:
Maybe the DS4 profile tries to access axis numbers or button numbers that are unavailable and that breaks Retroarchs alternative keyboard input for player 1?

Kivutar:
This problem is going to be difficult to solve. If we keep you changes, it will crash on all older kernels (and in Lakka we use a lot of old kernels). If we don't keep your changes, DS4 won't work on Ubuntu which is the distro where everything should work out of the box.

Schroedingers-Cat:
Sooner or later, v4.10+ will be used in any recent distribution, not just Ubuntu. 
RA needs to have kernel-dependent profiles (or something like that), otherwise it won't work everywhere out of the box.

twinaphex:
This keeps breaking my DS4 gamepad right now on Linux kernel (kernel 4.9.11).

Upstream Linux really made a goddamn mess out of this. We need a solution to this. Can any of you guarantee me that if I upgrade to a newer kernel, that the DS4 pad works properly again? I was burned before on a previous kernel which led to me reverting to 4.9.11 again.

Kivutar:
Even if it works with newer kernel, it will remain a problem for people using old kernels (which is very common in the embedded world)

Schroedingers-Cat:
Yes, that is why kernel-dependent profiles are the only way to solve this.

PR #368, opened 2017-08-18
raphnet technologies N64/Gamecube to USB adapter

nijoakim:


udev/raphnet.net_GC_N64_to_USB_Adapter_2_3.cfg
```diff
@@ -0,0 +1,26 @@
+input_device = "raphnet.net GC/N64 to USB, v2.3"
+input_driver = "udev"
```
RobLoach: Suggest adding three items:

- `input_device_display_name` for a pretty name
- `input_vendor_id`, this can usually be found when it's plugged in and you start RetroArch
- `input_product_id`, same as above

This will help RetroArch identify the device.

nijoakim:
Thanks for the feedback! I will add those probably on Monday as I'm away over the weekend.

udev/raphnet.net_GC_N64_to_USB_Adapter_2_3.cfg
nijoakim: Thanks for the feedback! I will add those probably on Monday as I'm away over the weekend.

RobLoach: Cool! If you switch to the SDL input driver, and re-run RetroArch, I believe the SDL input driver announces what it sees.

nijoakim: Now it's done! Let me know if you want me to change anything else.

paalfe:
I recommend not adding this, the button layout/assignment is different when used with GameCube or N64. Not possible to have one config fort this adapter suitable for both controllers. I have tried with my v2.9 of this adapter... You find my configs here https://github.com/paalfe/mixedcontent/tree/master/gaming%20(linux)/RetroArch/joypad-autoconfig_paalfe

nijoakim:
Ah! Right! I didn't think of that. I have only been using with GameCube controllers.

gre6102:
Question I have a raphnet ps3 pc joystick trying to figure out how to hook it to RetroPie if I go into retro arch it gives a code on the bottom 6017/2710. Any help would be greatly appreciated

udev/raphnet.net_GC_N64_to_USB_Adapter_2_3.cfg
```diff
@@ -0,0 +1,29 @@
+input_device = "raphnet.net GC/N64 to USB, v2.3"
+input_device_display_name = raphnet Gamecube/Nintendo64 to USB adapter.
```
RobLoach: Standard is to have quotes around the display name, without the period.

nijoakim:
@RobLoach: I added your suggestion. However, I thought this wasn't going to be merged anyways because of the reason @paalfe pointed out.

RobLoach:
While it could be problematic to have TWO configurations for the same thing, I don't even see ONE with...

```
input_vendor_id = 10395
input_product_id = 4
```

I feel like having at least one in there would be a good start. If people want to switch it out for a different config, then they could remap it.... Perhaps target the Gamecube controller config with this one? Let them remap it themselves if they want the N64 one?

nijoakim:
Okay! Sounds good to me! :)

PR #373, opened 2017-09-05
Added support for MAD CATZ Xbox-like controller

bassdr:


hizzlekizzle:
Looks good so far, but would you mind adding the button labels, as well? That will make it show the buttons' actual names in the core input remapping menu.

bassdr:
Sure, I did not know that was possible. I took the same labels as Microsoft_X-Box_pad_v2_US.cfg since this is the same controller.

hizzlekizzle:
Perfect. Thanks!

PR #378, opened 2017-10-21
Create Ultimarc_J-PAC.cfg

ryanfb:
See: https://www.ultimarc.com/jpac2.html


hizzlekizzle:
oh interesting. I didn't realize we could even map keyboard inputs to autoconfigs. Neat!

PR #402, opened 2017-11-29
Andrewchumchal patch 1

andrewchumchal:
Add Two New PlayStation 3 style controllers


hizzlekizzle:
Hi, it looks like we already have one for the MY-POWER 2In1:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/e2c6df1eb42cfd9796d90a2b9879b9a72c3d5d5d/udev/MY-POWER_CO__LTD__2In1_USB_Joystick.cfg

Would you mind adding the button labels to the BDA Pro Ex?

BDA Pro Ex.cfg
```diff
@@ -0,0 +1,28 @@
+input_driver = "udev"
+input_device = "BDA Pro Ex"
```
RobLoach: Looks like this is in the root directory? I believe this should be in `udev/` instead.

hizzlekizzle: oh yeah, good catch. Definitely needs to be in udev dir

MatiasFernandez:
@hizzlekizzle this has been open for almost 2 years. The MY-POWER profile is a duplicate, so we can skip it. The BDA Pro Ex isn't, so we may want to include it in the repo. Given that @andrewchumchal called them "PS3-like controllers" we may guess the labels for the buttons or maybe just add it to the repo. I think this won't be the first profile that is included in the repo without having labels.

I can prepare a PR for that if you want. If it doesn't make sense, we may just close the PR

offalynne:
> I think this won't be the first profile that is included in the repo without having labels
certainly not, haha 😄

looks like it is indeed a ps3 style pad:
https://itch.io/post/628320

hizzlekizzle:
> I can prepare a PR for that if you want.

Sure

> I think this won't be the first profile that is included in the repo without having labels.

Yeah, a very large number of profiles were added before button labels were even a thing. I have tried to make them a requirement since they were added, but it's possible some have slipped through the cracks :)

offalynne:
The mapping is [corroborated here] (https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt#L503) as the SDL format is Xbox/Sega style as opposed to RA's Nintendo style (ABXY instead of BAYX), and they all line up, so it's safe to assume this is correctly mapped and needs only the PS3 button labels added for completion.

hizzlekizzle:
closed by #573

PR #403, opened 2017-12-03
udev: Revert the Sony-PlayStation3-DualShock3-Controller-USB.cfg

orbea:
I originally submitted this `.cfg` a while ago and since then someone completely changed and broke it. It breaks even my keyboard! This took me a long time to understand....


twinaphex:
On which Linux kernel is this? IIRC at a certain point the kernel's DS3 gamepad mappings changed and this caused us to have to update the autoconfig file.

Maybe we could have a Linux kernel version check though in autoconfig and then determine based on that which autoconfig file (if there are more than one file) is ideal.

orbea:
I'm on 4.14.2. I recall this broke around 4.11 or 4.12.

orbea:
Yea if this is a problem in the kernel changing mappings a kernel check could help, but this makes me wonder why they changed it in the kernel in the first place?

Regardless there is something wrong with the current config file, it shouldn't break my keyboard.

orbea:
This is a related issue which I can still reproduce.
https://github.com/libretro/RetroArch/issues/4335

PR #404, opened 2017-12-07
Add Genius Maxfire G12U

alfrix:


hizzlekizzle:
This looks good, but could I trouble you to add the button labels while you're at it? That makes it easier to remap the buttons through the core 'controls' menu.

Issue #376, opened 2017-10-06
Microsoft_X-Box_One_S_pad_v2.cfg is certainly invalid


Alex131089:
Due to https://github.com/libretro/RetroArch/issues/3903, the udev/Microsoft_X-Box_One_S_pad_v2.cfg configuration is probably invalid, the file VID&PID is the one from ASUS (http://www.the-sz.com/products/usbid/index.php?v=0xB05&p=17CB).
Example : https://gist.github.com/Alex131089/2d8418a27164e78b1cf5b5881680332c
2578:1 (http://www.the-sz.com/products/usbid/index.php?v=0x0A12) reported instead of 1118:765 (Xbox One S using BT, reported as 746 using USB)

hizzlekizzle:
Oh no, that's a big problem! Do you have a correct vid/pid for it?

Alex131089:
All I can tell, is that the BT vid:pid for the corresponding USB vid:pid 1118:746 is 1118:765 (which is probably what was trying to be achieved).

However, between https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/Microsoft_X-Box_One_S_pad.cfg and https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/Microsoft_X-Box_One_S_pad_v2.cfg the configuration is not the same, so it's maybe another controller, or the BT mapping is different from the USB one (I don't believe it is), or it is another/custom configuration (more likely ?).

duganchen:
The values are vid:1118,pid:736. I was the one who added that profile, and those are the values for that controller (I just checked).

This issue can be resolved now, btw. It was fixed with the merge of https://github.com/libretro/retroarch-joypad-autoconfig/pull/409.

PR #409, opened 2018-01-02
Fix issues blocking DualShock 4 (CUH-ZCT2U) over BT on Linux

duganchen:
I started a forum thread about this issue here:

https://forums.libretro.com/t/wrong-joypad-profile-is-being-loaded/14009/2

As for what's in the pull request:

The vid and pid for the Xbox One S and Switch Pro controller profiles are fixed. I was the one who originally submitted these, and I've retested the profiles with the same controllers.

I also removed an old DS4 BT controller profile that, AFAICT, seems to be wrong and blocking the controller from working. My controller starts working well after that profile is removed, and a properly working profile is automatically loaded instead. It does not work with that profile present. (What likely happened: that profile is quite old, I'm on a 4.14 kernel, and the mappings have changed since then).

This also address https://github.com/libretro/retroarch-joypad-autoconfig/issues/376


duganchen:
This also address the following issue: https://github.com/libretro/retroarch-joypad-autoconfig/issues/376

PR #413, opened 2018-01-04
NEO 2in1 NV-PS3008

alfrix:
Tagging @hizzlekizzle @twinaphex because this gamepad uses the same VID and PID as DUALSHOCK-3
and it's recognized as one, but seems to have different DPAD settings, i don't have a DUALSHOCK-3 to test if the DS3 xinput-config is wrong, or this is just different, and if it is, will this break DS3?


alfrix:
@tony971 can you confirm that adding this config file does not break dualshock3 on xinput?

(tagging because you made the original DS3 xinput config)

tony971:
Not quite. The file already existed. I just swept through the xinput folder, standardizing formats. Sorry I can't help.

alfrix:
Did some tests changing the 'input_device' field, and only applies the config if its a match, so everything is fine, ready to merge

alfrix:
@fr500 since you have a DS3 could you please test on xinput (windows) if having this doesn't mess up the config (this pad uses same PID and VID) it appears to me that it doesn't.

i have seen:
https://forums.libretro.com/t/retroarch-menu-button-unconfigurable-by-design/14206/14
and noticed that the PS button isn't configured on xinput so it should be added to the autoconfig

PR #416, opened 2018-01-04
Create 8Bitdo_Pro_SN30_USB.cfg

paalfe:


hizzlekizzle:
Thanks for these!

paalfe:
My pleasure :) Will soon add these for android and xinput. Will add 8Bitdo SF30 Pro and 8Bitdo SF30 2.4G when they arrive... :+1:

PR #419, opened 2018-01-05
New android autoconfig

ZacharyFoxx:
iPega 9028 autoconfig to Android.


hizzlekizzle:
That looks good, but while we have you, can you add the button labels while you're at it? that helps with the input remapping menu.

ZacharyFoxx:
Hi,

How can I do this without open another pull request?

hizzlekizzle:
Very close, but I think the face button labels need to be swapped X<->Y and B<->A, based on an image I found of the pad online. I'll go ahead and merge it and just do it for you, though.

Thanks for the submission!

PR #431, opened 2018-02-07
redefinition of event codes for kernel > 4.15

tjustino:
Maybe the same fix that https://github.com/libretro/retroarch-joypad-autoconfig/pull/403


hizzlekizzle:
Thanks for submitting! We may hold onto this for a bit until the newer kernel filters down to more distros.

twinaphex:
Man, how many times are these guys going to screw around with the button bindings like this?

They already broke things badly one time before, now again.

orbea:
This is a duplicate of https://github.com/libretro/retroarch-joypad-autoconfig/pull/403.

hizzlekizzle:
Ok, we'll probably want to update this pretty soon. I think once Ubuntu puts out 18.04 (an LTS release), it'll be using 4.15 kernel.

orbea:
Sounds reasonable. :)

hizzlekizzle:
Ok, I merged the other almost-identical PR, so I'm going to close this one. Thanks for submitting, tjustino!

PR #435, opened 2018-02-21
Added 8Bitdo SF30 Pro controller in XInput mode

markand:
While there is already 8Bitdo_Pro_SF30_BT.cfg, this file is only for
DInput mode. Add this for XInput as well.

I hope this won't conflict with DInput mode but the vendor/product id are different so it should work.


markand:
Wait before merging, for some reasons on my other computer it gets the DInput file one instead.

markand:
You can safely merge this PR now.

PR #442, opened 2018-04-05
udev: Add SHANWAN PS3 GamePad.cfg

orbea:
A Chinese Dualshock3 clone.

Note: This might not work with older kernels?


hizzlekizzle:
I don't think we can merge this, as there are already a bunch of pads with the same vid/pid

orbea:
@hizzlekizzle I was aware of that, I have one of them.

However the device name is different from all the others and it works as a result without conflicting with the other gamepads.

Since there are multiple other configs with the same vid/pid, but different device names, every time I plug in this gamepad without this config all input (including keyboard) is entirely blocked.

orbea:
Anyways, it doesn't matter too much to me, I wasn't trying to buy a chinese knockoff and I'm going to try to return it as it was advertised as made by Sony...

The goal here was to add a missing gamepad config since I had the chance.

orbea:
Also linking this PR since the config is identical besides the device name.
https://github.com/libretro/retroarch-joypad-autoconfig/pull/403

PR #446, opened 2018-04-26
hid: Add 8Bitdo SF30 Pro

tsoliman:
Tested on macOS via USB on Android mode


RobLoach:
Looks great! Mind adding the label names?

Could likely copy them from https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/8Bitdo_Pro_SF30_USB.cfg#L35-L55

tsoliman:
Done. I've added comments as well.

tsoliman:
side observations about the 4 modes the device has:
- The default "Android mode" works but doesn't stay connected to BT on macOS for more than a few seconds (USB works fine with this PR)
- The "macOS mode" makes the controller pretend to be the "SCE Wireless Controller" in both BT and USB (and works fine)
- The "Windows mode" is basically xinput (USB is fine as a "Wired 360 controller" but BT goes nuts)
- The "Switch mode" makes the controller pretend to be a Nintendo Switch Pro Controller. BT works fine but USB registers another non-functional Switch Pro Controller (none of the buttons do anything). This is the only profile where connecting the USB doesn't auto disconnect BT. 

It is a weird device. Literally spoofing vendorid/deviceid like that is something I haven't seen before.

PR #447, opened 2018-04-29
hid: Add Mayflash Wii Classic Controller Adapter

tsoliman:
This USB controller adapter has 2 ports and presents itself as 2 devices to Retroarch.
There seems to be a race on which port gets to be device `#0` and which gets to be device `#1`

http://www.mayflash.com/products/pcusb/pc052.html


hizzlekizzle:
Looks great. Thanks!

Issue #454, opened 2018-05-29
fix the dragon rise zero delay encoder  across platforms for auto config


grant2258:
![dragonrise zero delay](https://user-images.githubusercontent.com/6128601/40644802-31c5f400-631c-11e8-826e-d7e2e22b7ef7.png)

all other buttons are mapped the same accept across these two configs accept

xinput 
input_l_btn = "6"
input_r_btn = "7"
input_l2_btn = "4"
input_r2_btn = "5"

udev
input_l_btn = "4"
input_r_btn = "5"
input_l2_btn = "6"
input_r2_btn = "7"

I used the udev config defaults they are shown in green in the picture the numbers are the physical joystick numbers in the picture as you can see these different button settings will change layouts on a arcade panel with these defaults not being the same.  This encoder only supports a left analog stick as well and this is the mode it defaults to you have to press the mode button to change between digital and analog. Hope this clears up some confusion on this controller and how to set the config up.

input_device_display_name = "DragonRise zero delay encoder"

input_x_btn = "0"
input_a_btn = "1"
input_b_btn = "2"
input_y_btn = "3"
input_l_btn = "4"
input_r_btn = "5"
input_l2_btn = "6"
input_r2_btn = "7"
input_select_btn = "8"
input_start_btn = "9"
input_l3_btn = "10"
input_r3_btn = "11"

input_left_btn = "h0left"
input_right_btn = "h0right"
input_up_btn = "h0up"
input_down_btn = "h0down"
input_l_x_plus_axis = "+0"
input_l_x_minus_axis = "-0"
input_l_y_plus_axis = "+1"
input_l_y_minus_axis = "-1"

#########

input_x_btn_label = "X"
input_a_btn_label = "A"
input_b_btn_label = "B"
input_y_btn_label = "Y"
input_l_btn_label = "L"
input_r_btn_label = "R"
input_l2_btn_label = "L2"
input_r2_btn_label = "R2"
input_select_btn_label = "Select"
input_start_btn_label = "Start"
input_l3_btn_label = "L3"
input_r3_btn_label = "R3"

input_left_btn_label = "D-Pad Left"
input_right_btn_label = "D-Pad Right"
input_up_btn_label = "D-Pad Up"
input_down_btn_label = "D-Pad Down"
input_l_x_plus_axis_label = "Left Analog X+"
input_l_x_minus_axis_label = "Left Analog X-"
input_l_y_plus_axis_label = "Left Analog Y+"
[...]

PR #455, opened 2018-05-30
fix up this encoder across platforms

grant2258:
please see https://github.com/libretro/retroarch-joypad-autoconfig/issues/454 for details


hizzlekizzle:
I guess this will interfere with some of the Dragonrise gamepads?

grant2258:
no they are a diff  number and its numbered the same as the dual shock one anyway the only conflict is the retro n64 posted on another issue here

grant2258:
if there is any issue or conficts some how ill add the second analogue stick in :). I can see there being any though

dankcushions:
why did you remove the instructions for how to set up this stick re: R1/R2 third column layout? https://github.com/libretro/retroarch-joypad-autoconfig/pull/455/files#diff-2fcd714f4f167d953135c36721912e30

that was good information. reinstate it.

PR #457, opened 2018-06-03
Add support for YSTEK MICREAL USB Gamepad

disconn3ct:
This is a gamepad on Amazon, using the same generic receiver as everyone
else. It ships with NES/SNES classic receivers as well as USB but the
existing mappings cause it to scroll madly as soon as it is plugged in.


hizzlekizzle:
Can you add the button labels, as well?

disconn3ct:
Labels are added

hizzlekizzle:
Looks great. thanks!

PR #464, opened 2018-06-17
Update raphnet.net_4nes4snes.cfg 

lucasmr:
Product ID was wrong with firmware 1.5 and mapping was incomplete.


hizzlekizzle:
Is there a reason this was closed? Are you planning to open another PR, perhaps?

lucasmr:
Yes, I was thinking about the hybrid nature of this adapter. I did not test the adapter with a NES controller, so I don't know if the mappings would be same for the A, B, Start and Select buttons. The other SNES buttons wouldn't matter because they are not present, obviously.

PR #466, opened 2018-07-01
Create HuiJia  JC-W01U.cfg

Hiki1:
Added a configuration file for Wii controller game pad to USB converter


udev/HuiJia  JC-W01U.cfg
```diff
@@ -0,0 +1,50 @@
+input_driver = "udev"
+input_device = "HuiJia  JC-W01U"
```
RobLoach: Mind adding a `input_device_display_name` too?

Hiki1: I just added it. Thanks for noticing!

PR #472, opened 2018-08-18
adding GameStick Controller config

jmandawg:
People might have this controller laying around like me, surprisingly it works good, but no R2/L2 buttons, also the select key is labelled back and sends a keyboard "backspace" so that needs to be mapped manually.  Also attached is the udev rule.  I can create a PR for that as well, but not sure where it goes (had to rename to .txt to attach).

[99-gamestick-controller.rules.txt](https://github.com/libretro/retroarch-joypad-autoconfig/files/2299096/99-gamestick-controller.rules.txt)

### DIFF
--- a/udev/GameStick Controller.cfg
+++ b/udev/GameStick Controller.cfg
@@ -1,0 +1,68 @@
+#Gamestick bluetooth controller
+
+# Note: The select button which is labelled "back" on the controller needs to 
+# be manually set in retroarch ui as it actually sends a keyboard 
+# "backspace" key code
+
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
that looks good, but would you mind adding the button labels, as well, while you're here? As for the udev rule, we don't really have anywhere to store such a thing, but you could maybe paste it into the autoconfig profile as a comment.

jmandawg:
how's that look?

hizzlekizzle:
eyyy, looks beautiful. Thanks!

PR #473, opened 2018-08-25
Adding 8Bitdo modkit configs.

Krispy3000:
MD30 is the only USB config because it's the only controller with an exposed USB port.


hizzlekizzle:
those look great! thanks!

PR #477, opened 2018-09-19
Xbox One S Wireless Controller support for Android

dekomote:
Added correct mapping for Xbox One S Bluetooth Controller for android system, tested on retroarch + android TV 6. The current mapping is for the original Xbox One Controller and it's incomplete


hizzlekizzle:
Looks good! Thanks!

bparker06:
It's also worth noting that Android Pie (9.0) includes a mapping for this pad, not sure if it conflicts or would work any differently with the default profile though.

https://www.xda-developers.com/android-pie-controller-mapping-fix-xbox-one-s/

dekomote:
@bparker06 It won't cause a conflict. So far the pad used the configuration for the original Xbox One controller, which has a bit different mapping than the S. Only 5 buttons were out of place.
The Pie update will help with android games, I don't think it'll change anything in retroarch.

On a semi-related note, is it possible for the autoconfig to define the input as retropad with analog?

PR #478, opened 2018-09-21
Raphnet tech autoconfigs

raphnet:
Add autoconfig files for raphnet-tech Classic controller to USB and SNES controller to USB adapters.


hizzlekizzle:
those look great, thanks!

PR #480, opened 2018-09-26
Dualshock: Remove the old file, append a .cfg in the new

RobLoach:
Seems there are duplicate vendor and product ids over in this one:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/Sony-PlayStation4-DualShock4v2-Controller.cfg

Also deleted an old one that also didn't have a .cfg extension. Since this results in two .cfg having he same vendor and product ids which one should we use?


SapphireTactician:
Interesting. It looks like everything is identical except for the fact that I feel mine is better sorted / organized whereas the previous file used the default sorting.

Question: If an existing configuration with identical settings already existed, then why didn't RA map mine correctly then? Was it because it was missing the extension? (*I made the same omission by mistake when copying the naming style of those*)

hizzlekizzle:
I *think* it's okay to have duplicate vid/pid, as long as the name is different. That is, I think the name figures into the scoring system. @fr500 may have more to say on it.

SapphireTactician:
That wasn't the default `input_device` name though, the default name of mine matches the one in the config file linked in the first post / comment. However, since the inputs were mismatched on RA, I figured I'd rename it to prevent messing with any other config files.

I'm hoping that's ok, and that I didn't foul up.

MatiasFernandez:
@hizzlekizzle #571 was an alternative to this PR so I think we can close this one

PR #482, opened 2018-09-29
Create GPD Win 2 X-Box Controller.cfg

chris062689:
Added support for the GPD Win 2.

This was tested with kernel 4.18.0 which added supported for switching the GPD Win 2 to controller mode.


hizzlekizzle:
This looks good so far, but would you mind adding the button labels, as well? That will make the buttons have the correct names in the controls/remap menu.

hizzlekizzle:
@chris062689 Not exactly, but you can look at most/any of the other autoconfigs and see how it's done. It's just a matter of putting a name to the buttons on a new line. For example, input_b_btn would get input_b_btn_label = "Whatever"

MatiasFernandez:
There is already a profile for this device in the repo, including labels. I think we can close this PR @hizzlekizzle

PR #485, opened 2018-11-08
Create HORI CO.,LTD. POKKEN CONTROLLER.cfg

j1dopeman:


hizzlekizzle:
This looks great, thanks!

But, while we have you, could we trouble you to add the button labels, as well?

j1dopeman:
Okay done. There are 2 other start/select like buttons on the face that are labeled ZL/ZR. I made ZR a menu button for convenience.

hizzlekizzle:
Seems good to me :)

Thanks!

PR #486, opened 2018-11-16
Add the 8bitdo SN30 Bluetooth Gamepad (VID 1118 / PID 736)

Dwedit:
I created a profile for the 8Bitdo SN30 Bluetooth Gamepad, which is Vendor 1118 / Product 736, unlike the other profile.  I named it "_ALT".


Dwedit:
Closing the PR because it seems the gamepad is spoofing a real XBOX 360 controller.

PR #487, opened 2018-11-19
Create Xperia_Play_zeus2.cfg

Royerson:
device name for CM9/10


Royerson:
Actually, looking at [android_input.c](https://github.com/libretro/RetroArch/blob/f318f4a1a99b258d1bb45bcdaf136ed83809c855/input/drivers/android_input.c#L1079) it seems like the all caps XPERIA is caused there instead of by cyanogenmod. If so would probably be cleaner to change it there instead?

PR #488, opened 2018-11-25
Create autoconfig for 3h dual arcade jamma to usb adapter

0x15e:
Autoconfig for a JAMMA to USB input adapter board


hizzlekizzle:
Looks great, very informative comments. Thanks!

PR #491, opened 2018-12-06
Add udev configuration for PlayStation Classic USB controller

andshrew:
This is a udev configuration for the PlayStation Classic USB controller. Tested on Retroarch running on a Raspberry Pi 3.


hizzlekizzle:
Looks great. Thanks again :)

PR #492, opened 2018-12-06
Add android configuration for PlayStation Classic USB controller

andshrew:
This is tested on Retroarch running on a Shield TV.

Could I also check if it is preferable for me to submit these new configurations as individual pull requests, or would you prefer them bundled into one request?


hizzlekizzle:
We'll take them either way. I guess other things being equal, one big PR is preferable, but it's not a big deal either way.

andshrew:
Thanks, I'll keep that in mind for the future.

PR #495, opened 2019-01-07
New xinput configs for 8bitdo Gbros Adapter

andshrew:
These are two xinput configurations for the 8bitdo Gbros. Adapter. This was made using a device on the latest firmware 2.20 which changes r2/l2 to axis.

This adapter lets you connect Gamecube or controllers with the Wii remote adapter (Classic Controller, Classic SNES, Classic NES etc.). Unfortunately it will always present itself as the same device regardless of what is connected, so I have created the labels based on a Wii Classic Controller but the bindings should work regardless of the controller you use.


hizzlekizzle:
Looks great. Thanks!

PR #504, opened 2019-01-14
Added 8BitDo SN30 GP for xinput, android, and udev

Tony556:
I hope I did this right!

Before it would try to use "Retro-Bit_8Bitdo_RB8-64" which is very inaccurate, and didn't even have the same PID (thank goodness)

Product Page:
http://www.8bitdo.com/sn30-gp/


hizzlekizzle:
These look great, yeah! Thanks!

PR #507, opened 2019-01-27
Create RetroFlag_Classic_USB_Controller-M.cfg

max-besch:


hizzlekizzle:
This looks good, but the vid/pid conflict with the Buffalo USB SNES pad.

max-besch:
RetroFlag "Classic USB Controller-M" and "iBuffalo Classic SNES USB Gamepad (BSGP801)" have different device names (input_device) so there is should be no conflicts in matching since algorithm uses all three VID, PID and device name (input_device)

hizzlekizzle:
Alright, after consulting with some other folks, I'll go ahead and merge it and we can revert it if it causes a bunch of false-positives.

PR #510, opened 2019-02-09
hid: Add 8Bitdo NES30 Pro

cold-brewed:
Add support for 8Bitdo NES30 Pro. 

Notes:

* This is for the Android mode where you power on with no additional buttons pressed. It shows up as "8Bitdo NES30 Pro" this way.
* Works for both BT and USB
* Tested on the latest firmware as of now (4.10) as well as 4.01
* You can also start it up in macOS mode by holding A during power on, but then it spoofs as a DS4 controller and this config won't apply
  * If you run it in this mode it makes RA segafult when exiting so I recommend running it in the default (Android) mode


hizzlekizzle:
Looks good, thanks!

PR #512, opened 2019-02-12
hid: Add 8Bitdo SN30 Pro

cold-brewed:
Add support for 8Bitdo SN30 Pro.

Notes:

* This is for the Android mode where you power on with no additional buttons pressed. It shows up as "8Bitdo SN30 Pro" this way.
* Works for both BT and USB
* Tested on the latest firmware as of now (1.29)


hizzlekizzle:
Looks great! thanks again!

PR #514, opened 2019-02-21
Added autoconfigs for udev for 3 controllers

mrmatteastwood:
Hey, Libretro team! I own 2 Afterglow Xbox 360 controllers that got recognized as 3 different devices on my PC. In all 3 instances, I made separate autoconfig files for them while using the udev driver. Here they are.


hizzlekizzle:
Hmm, it seems we have a conflicting profile already with different mapping:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/0ebca0c491992dbbcdb5f960e261236c547b0ba3/udev/Afterglow_Gamepad_for_Xbox_360.cfg

Is this one incorrect?

mrmatteastwood:
Hmm, I can't say for sure. When I plug in my Afterglow controller, that autoconfig isn't loaded (fyi, I'm on Linux Mint 19). That'S why I ended up making the new one. Perhaps the fact that the pre-existing file has underscores in the filename while mine doesn't makes a difference? I can say for sure that my new file works with my controller.

MatiasFernandez:
@hizzlekizzle Afterglow controller has a conflict, and I think we don't have the tools to understand which of those is the right one (the one in the repo, or the one proposed in the PR). The one in the repo has one more button mapped, so I feel it could be a bit "better". Should we split this PR to just add the ones that are new and skip the one that is conflicting?

I can add some notes to the README explaining that if you have a duplicated profile you won't be able to upload it to the repo and you just need to generate your own profile and store it yourself.

hizzlekizzle:
yes, seems like a good course of action to me.

hizzlekizzle:
closed by #573

PR #516, opened 2019-03-03
Update 8Bitdo_N30_Modkit_BT.cfg

dauntless360:
Received a new 8Bitdo N30 Modkit (NES Classic Controller Version) with the latest firmware (v6.10) and it didn't work out of the box with retroarch's autoconfig. I had to create a new autoconfig file to work with the updated firmware. Both the VID and PID changed. The up/down axis values also had to be inverted. The new changes work with the latest current firmware and were tested with the android version of retroarch on the Shield TV.


hizzlekizzle:
Ah, this is good! The previous vid/pid were causing conflicts with 360 pads. This should fix that right up.

PR #520, opened 2019-03-10
Autoconfig file for Gametel Bluetooth controller

tvalenta1:
This is an autoconfig file for my Gametel Bluetooth controller (www.gametel.se).

### DIFF
--- a/udev/Gametel_Bluetooth_controller.cfg
+++ b/udev/Gametel_Bluetooth_controller.cfg
@@ -1,0 +1,28 @@
+input_driver = "udev"
+input_device = "Gametel"
+input_device_display_name = "Gametel Bluetooth controller"
+#On Linux - if this Bluetooth controller is not recognized by the system then create a file "99-gametel.rules" with following content:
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
That looks good so far, but can we trouble you to add the button labels while you're at it?

tvalenta1:
Sure. No worries. I will add that.
Just not sure what to add for the "action" buttons... as there are no labels on the controller except for symbols "o I", "I o".
input_b_btn_label = "-o-"
input_y_btn_label = "oI"
input_a_btn_label = "Io"
input_x_btn_label = "_o_"
Will this be ok? Or should I use some generic words (left button, right button)? Or labels for these buttons?
01a1bf6f2dc923873a31afe99e23af14-480-80

hizzlekizzle:
That's up to you. I think something like left/right/top/bottom is fine.

twinaphex:
@hizzlekizzle Maybe add the missing labels to it if all it takes is just looking at a screenshot? That way we can merge this. This PR has stalled otherwise.

PR #527, opened 2019-04-03
Add autoconfig for ipega PG-9087 controller

fasheng:
ipega PG-9087 could connect to PC/Android with bluetooth or wired USB, for bluetooth, its device name is "PG-9087", and for USB it just actor as "ShanWan     PC/PS3/Android Gamepad".

http://ipega.hk/gamehandle/53-97.html


hizzlekizzle:
These look good, thanks! Is there a reason some of them don't have the button labels?

fasheng:
Yes, just not sure the label names for ShanWan before, now I copy them from `udev/SHANWAN PS3 GamePad.cfg` and update PG-8097's labels, too.

Please review again.

DartPower:
My iPega PG-9087s detecting as Ipega 9023

fasheng:
Well, main is PG-9087 not PG-9087s, and both work OK for Linux and Android.

Here is the official page for 9087s, so they are two different modes in fact:
http://ipega.hk/gamehandle/53-194.html

I guess there buttons should be similar, so you could edit the `input_product_id` in config file and test in local. Good luck!

PR #528, opened 2019-04-03
udev: added Gioteck VX 2 controller

5schatten:


hizzlekizzle:
We try to avoid putting hotkey assignments into the autoconfigs, since they're very personal and controversial. The only ones we typically allow are for home/menu buttons, so would you mind removing the others?

Also, can you add button labels?

5schatten:
I can delete the hotkeys but what about button labels? Well I don't own this controller -> someone who uses my LibreELEC fork asked me if I could add this file to make it work out-of-the box but I guess it would be nice it works with upstream.

PR #529, opened 2019-04-03
udev: added DS4 Jet Black V2 controller

5schatten:


julianxhokaxhiu:
This is redundant. There is already a config for this pad. See https://github.com/libretro/retroarch-joypad-autoconfig/pull/538

5schatten:
Well I got reports that the config did not work in the past and asked a user for a working one. I'm closing the PR if the config is fixed now.

PR #533, opened 2019-04-27
Playstation classic like chinese gamepad

h1aji:
This is a Playstation style gamepad that uses a generic chinese USB gamepad encoder chip common in other gamepads with different button configurations.


h1aji:
Link to device: https://www.aliexpress.com/item/USB-1-01-2-0-Controller-Gamepad-for-PC-USB-Joystick-for-PC-Game-Wired-Computer/32952321160.html

hizzlekizzle:
This directly conflicts with a few other generic pads of different styles, but the names are different, so they *shouldn't* interfere with one another.

PR #536, opened 2019-05-22
Add 8BitDo M30 Bluetooth Controller hid config

ZackXu0:


hizzlekizzle:
While we have you, can I trouble you to add the button labels, as well?

ZackXu0:
Ok, I will try to re-edit these configurations.

PR #543, opened 2019-07-09
Added xinput autoconfigs for 8Bitdo M30 controller

DeusMalsith:
Configs include hat, left joystick, and right joystick associations for D-Pad.


DeusMalsith:
Added specificity: The autoconfigs are for USB and Bluetooth.

hizzlekizzle:
Looks great. Thanks!

PR #554, opened 2019-08-17
Added Retro-Bit Tribute64 - USB (D-Input)

andrewkendall:
I've added out-of-the-box support for the Retro-Bit Tribute64 USB controller (http://retro-bit.com/tribute64).

The controller defaults to what Retro-Bit call D-Input, although when saving an Autoconfig Retroarch uses input_driver = "xinput", so this is what I've used.

I've fully tested against Mupen64Plus-Next and everything maps correctly.


hizzlekizzle:
Looks good! Nice to see the duped a-button trick implemented :)

andrewkendall:
Thank you! 😅 I'm pulling correct values from device manager, doing some testing and will resubmit.

PR #555, opened 2019-08-17
Updated Retro-Bit Tribute64 - USB (D-Input).cfg

andrewkendall:
Updated input_vendor_id and input_product_id for Retro-Bit Tribute64 - USB (D-Input)

I've now taken the vid/pid from Windows device manager rather than the auto-generated RetroArch values.

Tested and fully working in Mupen64Plus-Next, hopefully, these are the correct and unique device values.

### DIFF
--- a/xinput/Retro-Bit Tribute64 - USB (D-Input).cfg
+++ b/xinput/Retro-Bit Tribute64 - USB (D-Input).cfg
@@ -4,8 +4,8 @@
 input_driver = "xinput"
 input_device = "Controller (Dinput)"
 input_device_display_name = "Retro-Bit Tribute64 - USB (D-Input)"
-input_vendor_id = "1118"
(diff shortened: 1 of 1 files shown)


twinaphex:
Hmm, seems there is a merge conflict here that has to be resolved first.

hizzlekizzle:
Those are the same values, just in hex format, unfortunately.

andrewkendall:
Sorry about that! It looks like those must be the only values in D-Input mode so I'll close this. I'll look at the other mode and see if values are unique to the device for that one. Thank you for spotting the values were the same, just in hex. It's a shame as this is the default mode every time it's plugged in, but hopefully, the config still comes in useful to someone.

hizzlekizzle:
No problem! We appreciate the submission(s) all the same :)

PR #559, opened 2019-09-10
Update DUALSHOCK_4.cfg to use L2 and R2 as axes instead of buttons

MatiasFernandez:
Dualshock 4 has analog triggers (L2 and R2) but retroarch's autoconfig file sets them as buttons. Because of that, it's impossible to use the triggers as analog controls in games. Changing this to analog triggers is a better mapping of the physical controller, because they are actually axes and not buttons.


MatiasFernandez:
@twinaphex pinging you because you were the last one touching this file

hizzlekizzle:
I couldn't pin down any testers for this, but it seems like a good idea to me, so I'll go ahead and merge it and if it causes issues, we'll just revert it :)

Thanks!

MatiasFernandez:
Thanks @hizzlekizzle I tested it with my own controller and it worked correctly. Let me know if you have any specific test in mind I can reproduce on my side

davidhedlund:
Can you please comment this link above your the variables in the autoconfig file, in order to make it easy for people to understand why the values are different from the values produce from the RetroArch GUI?:
* https://github.com/libretro/RetroArch/issues/6920

Issue #322, opened 2017-02-02
README and CONTRIBUTING information


CoalaJoe:
Would be great if the structure of this repo would be explained.

hizzlekizzle:
Yeah, that's true. If you would be interested in writing something up, I'll merge it.

CoalaJoe:
I can try. But I don't know how or/and where to add what config file.

manuti:
How I can add the .cfg file for a new device (Gioteck VX2 Wired Controller.cfg) is working in my Lakka on Orange Pi and I want to share the configuration.
Thanks

gouchi:
@manuti : please read [this documentation](http://www.lakka.tv/doc/Contributing-your-joypad-config). Thank you.

MatiasFernandez:
@hizzlekizzle should this be closed?

PR #563, opened 2019-09-16
Propose README.md file including some useful documentation about the project

MatiasFernandez:
A while time ago @CoalaJoe created an [issue](https://github.com/libretro/retroarch-joypad-autoconfig/issues/322) with the idea of creating a README file for the project. @hizzlekizzle was positive about the idea but it seems nobody created a PR to deal with that.

Recently I've spent some time dealing with issues inside autoconfig profile files and I learned a bit about how things work. So I wanted to contribute to the project writing down some useful bits of documentation around autoconfig profiles.

I hope you find this useful, let me know what do you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,143 @@
+# Joypad Autoconfig Files
+
+This repository stores joypad autoconfig files for Retroarch. RetroArch is the reference frontend for the libretro API. 
+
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
I love it! Thanks for creating this!

offalynne:
a graphic for the readme that illustrates the RA virtual gamepad layout would be helpful since the BAYX layout can be a bit counterintuitive/ambiguous

PR #566, opened 2019-09-17
Move profile to the right path

MatiasFernandez:
Recently a new profile was added for a controller using udev driver but it was added to the root of the repository. AFAIK it should go to the udev folder

This PR is to move it there


hizzlekizzle:
whoops, good catch. thanks.

Issue #424, opened 2018-01-10
Conflicting configs DragonRise and Retrolink_N64


alfrix:
On both dinput and xinput

1. DragonRise_Inc.___Generic___USB__Joystick__
2. Retrolink_N64_USB

Have the same name VID and PID:
```
input_device = "Generic   USB  Joystick  "
input_vendor_id = "121"
input_product_id = "6"
```
The config files are very different, which means the gamepads are not the same.

Since RetroArch applies the last file name (in this case Retrolink_N64.cfg) as a workaround for DragonRise i have renamed the file as Z.cfg

MatiasFernandez:
I know this issue is old, but this comment help other people reading this issue. I have an alternative (not original) PS3 controller that also has a conflict with these other two examples you mentioned here.

My guess is that some generic controllers report themselves to the OS using a vendor ID, product ID and input device name that is pretty generic. It may be that some cheap generic circuits are being used for completely different layouts, so there's a naming colision here.

Retroarch uses three gamepad attributes to uniquely identify the controller but that is not enough. It will be great to have a way to remove the ambiguity by a setting in the UU (manual interaction) or maybe using other controller details to improve the matching process (like number of buttons and axes). 

I don't think the devs are super interested in doing something like this due to the low presence of such generic controllers within retroarch's user base. 

An easy workaround for people having this issue is to remove in their own environments the autoconfig file that creates the conflict.

offalynne:
It'd definitely be better to remove this VID/PID from autoconfig altogether. There are at least a dozen devices that use the chip, all of them have different layouts. Retrobit continues to use this chip in new devices including their recent Mega Drive and Saturn pads, many of which will be packed in with the forthcoming Mega Drive/Genesis mini consoles. To date there's no obvious consensus between device type and button layout, so identifying them with the existing system is much more likely to be wrong than right.

offalynne:
See also https://github.com/gabomdq/SDL_GameControllerDB/issues/202

hizzlekizzle:
How would you guys feel about me commenting the VID/PID on all of these and letting it match on just the name?

offalynne:
That would certainly before the best. Far as I can tell they all share the name ‘DragonRise Inc. PC TWIN SHOCK Gamepad’, so ID by name is likely to fail as well.

MatiasFernandez:
I think that won't fix the issue @hizzlekizzle 

[...]

PR #568, opened 2019-09-18
Disable conflicting (duplicated) profiles

MatiasFernandez:
As we discussed in #424 we need to clean up duplicated/conflicting profiles from the repo. 

The problem with duplicated profiles is that Retroarch isn't capable of disambiguating between them, so it will choose one or the other just by chance.  

The goal of this PR is to keep just one set of the duplicated profiles enabled, commenting `input_device`, `input_vendor_id`, and `input_product_id` lines to disable those profiles.

cc @hizzlekizzle

### DIFF
--- a/android/Defender_Game_Racer_Classic.cfg
+++ b/android/Defender_Game_Racer_Classic.cfg
@@ -1,7 +1,14 @@
 # Verify if D-pad works with hat controls
 
-input_device = "USB Gamepad"
 input_driver = "android"
+
+# Currently Retroarch can't differentiate between input devices
+# sharing the same input_vendor_id, input_product_id, and
+# input_device. We decided to disable the duplicates until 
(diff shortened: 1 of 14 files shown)


offalynne:
pardon more armchair opinion but perhaps it’d be useful to leave a comment referring to which config is left active (which the commented-out config defers to) ?  it’s easy enough to grep for the file with same vid/pid/name but leaving a note might make ease eventual resolution or further work to this end.

MatiasFernandez:
@offalynne I like the idea. I will update the PR with that :)

MatiasFernandez:
@offalynne done! I updated the commit with your suggestion

hizzlekizzle:
this looks good to me, and the note telling the remaining profile was a good idea. Thanks for handling this, @MatiasFernandez

PR #583, opened 2019-09-30
Add autoconfig for HOLTEK's Twin Shock

requeijaum:
image

Another DualShock rip-off for PC.
Originally in Brazilian market since 2004 under "Clone" OEM branding.
I need to test the analog mode, though.

On my controller: the analog mode PCB contact is broken. I might solder a push-button later.
The analog axis from left and right thumb are simple guesses from my usage with a Windows driver... long time ago.

-----

Edit: the D-pad usually broke a lot on these controllers. There was a recall but that didn't resolve anything. I still have mine and it's plastic is very strong.

Also: mind that the buttons are numbered starting from 0 and not 1 - as printed WRONGLY on the controller.


hizzlekizzle:
hey, thanks for this. is there a reason the button labels are all commented out?

requeijaum:
I preferred not to mess with the buttons labels since users can make a mess of it.
That's all because of the buttons markings wearing out and there is an offset of "1" between a marking and the real button (eg. button 0 is marked as 1).

offalynne:
the 0/1 difference isn’t a mistake; windows enumerates buttons in a “human readable” way in user-facing interfaces, ie 1-n instead of 0-n.

HOLTEK_Twin-Shock.cfg
```diff
@@ -0,0 +1,54 @@
...
+input_l_y_plus_axis = "+1"
+input_l_y_minus_axis = "-1"
+input_r_x_plus_axis = "+3"
+input_r_x_minus_axis = "-3"
+input_r_y_plus_axis = "+4"
+input_r_y_minus_axis = "-4"
+input_menu_toggle_btn = "8"
+input_b_btn_label = "3"
+input_y_btn_label = "4"
+#input_select_btn_label = "Select"
```
RobLoach, line 32: No need to comment these out, as mentioned above :wink:

RobLoach (changes requested):

HOLTEK_Twin-Shock.cfg
```diff
@@ -0,0 +1,54 @@
+input_device = "HOLTEK Twin Shock"
```
RobLoach, line 1: Would be good to have a `input_device_display_name` too.

offalynne:
@requeijaum how did you get the vendor and product fields here? did you author the cfg manually ? the autoconfig format uses decimal and these look to be hex values

requeijaum:
I used macOS System Profiler, Linux's lsusb and Windows 10's Device Manager.
VID and PID are hex values. I didn't know about autoconfig's decimal format.


Em sáb., 13 de mar. de 2021 às 19:05, alynne o. ***@***.***>
escreveu:

requeijaum:
Hey, @RobLoach 
What do you think? Are the USB VID and PID values OK ?
I will also approve your requested changes tomorrow.

offalynne:
assuming rob doesn’t have the hardware on hand to check. they are indeed supposed to be decimal, check other configs to compare

RobLoach:
Merging this, going to fix it up after the merge.

PR #584, opened 2019-10-01
Adds SEGA Genesis Mini 6-button 

evilwraith:
Official SEGA controller from JP console


evilwraith:
> vid & pid should be in decimal not hex

It is in decimal, just missing the zeroes from the front as quite a few of the config profiles do. Closing this as I have an error in the file anyway.

PR #588, opened 2019-10-07
Add config for moga pro

mikegr:
Added config for older Moga Pro controller
http://web.archive.org/web/20160308020701/http://www.mogaanywhere.com/controllers/moga-pro-controller/

Right analog did not work. Should I keep entries empty or remove them?


offalynne:
Looks like it's 
`input_r_x_plus_btn = "+2"
input_r_x_minus_btn = "-2"
input_r_y_plus_axis = "+3"
input_r_y_minus_axis = "-3"`
as per
https://github.com/spurious/SDL-mirror/blob/ace57ce610fbe1d0602012264101157c964f6b33/src/joystick/SDL_gamecontrollerdb.h#L453

mikegr:
I added this including labels.

PR #590, opened 2019-10-12
Update Logitech Cordless RumblePad 2.cfg

seriouslyunderappreciated:
Hey, I actually own a cordless rumblepad 2 and the autoconfig for the triggers and bumpers is incorrect.

This small change fixes the problem.

Also, this gamepad is listed as xinput when it's definitely dinput. The logitech f310 and logitech f710 are the xinput successors.


hizzlekizzle:
oh yeah, definitely backwards. Thanks for the fix :)

PR #591, opened 2019-10-14
Add config for Mayflash Arcade Fightstick F101

jstolarek:
What the title says.


hizzlekizzle:
Hmm, dunno why the continuous integration check isn't showing as finished... I'll merge it anyway. thanks!

PR #592, opened 2019-10-14
Fix swapped buttons for Logitech F310

jstolarek:
For whatever reason current autoconfig for Logitech F310 gamepad swaps A with B and X with Y - both when it comes to logical mappings and labels. This patch fixes that. The end result is exactly the same, except that things are now done in straightforward fashion.


jstolarek:
> For whatever reason

Ok, I think the actual reason for doing this is because RetroArch refers to buttons not only by their names A,B,X, and Y, but also by their placement up/down/left/right. So the mapping that already exists in the repo maps the buttons according to their physical locations on a pad (which results in swapping A with B, and X with Y) and then fixes that swap by also swapping labels. This is correct, but confusing. My mapping in the PR maps the buttons according to their labels on the pad, not on their physical placement, so RetroArch's B button is mapped to B button, even though it is the right button and not the bottom one.

hizzlekizzle:
I think the old way is correct (or at least: consistent with how we've been doing it). That is, the labels should match the physical controller's labels, which are based on the what the user sees when they look down at it, rather than the retropad labels, which are just based on location. In this case, the pad's labels are Xbox-style, which swap X/Y and B/A vs the retropad.

jstolarek:
> I think the old way is correct (or at least: consistent with how we've been doing it). 

In that case please feel free to discard this PR.

PR #595, opened 2019-10-18
add additional ports for Wii U GameCube Adapter

nrdxp:
Fix port 2-4 not being auto-configured by adding profile for each port.


hizzlekizzle:
Does this report a vid/pid? If so, it would be good to have that included in them, I think.

Also, could I trouble you to add button labels to these (and the original Port 1 file, if you don't mind)?

nrdxp:
Unfortunately I cannot get the vid/pid for the controllers only the adapter itself. I took care of the button labels. After using this configuration for the past 24 hours I realized that the Z button and triggers are poorly mapped, so I also changed them.

The Z button was previously mapped to 'select' which seems like kind of a waste, so I set it to R instead. I also moved the triggers from L and R, to L2 and R2.

PR #596, opened 2019-11-01
Add config for Mayflash Arcade Fightstick F101 XInput mode

bsdf:
the f101 has a toggle for xinput mode with different mappings


hizzlekizzle:
looks good, thanks!

PR #597, opened 2019-11-14
Create SWITCH CO.,LTD. USB Gamepad .cfg

donald-keighley:
Add retro-bit SEGA Genesis 8 Button Arcade Pad autoconfiguration.


offalynne:
kudos to retrobit for shipping yet another pad with an indistinguishable generic dragonrise board 🙄

udev/SWITCH CO.,LTD. USB Gamepad .cfg
```diff
@@ -0,0 +1,38 @@
...
+input_driver = "udev"
+input_device = "SWITCH CO.,LTD. USB Gamepad "
+input_vendor_id = "121"
+input_product_id = "17"
+input_device_display_name = "retro-bit SEGA Genesis 8-Button Arcade Pad"
+
+input_a_btn = "1"
+input_b_btn = "2"
+input_x_btn = "0"
+input_y_btn = "3"
```
miklosaubert, line 11: This will map the physical A, B, X and Y buttons on the controller directly to their similarly-labelled buttons on the Retropad. While this is straightforward in terms of button labelling (A is A, B is B...), it doesn't match the _physical layout_ of the Retropad.

We could argue that there's no such thing as a Retropad in the physical world, so we can map and remap as we please, but the Retroarch UI uses a layout of A, B, X and Y that is the same as the SNES controller (top=X, right=A, bottom=B, left=Y)

I personally think this should be reflected in this config file, i.e:
```
input_a_btn = "2"
input_b_btn = "1"
input_x_btn = "3"
input_y_btn = "0"
```
This has the following benefits:
- makes navigating the Retroarch UI easier ; of course the button labels are swapped, but the default on-screen help matches the physical layout on the pad, i.e. Retropad's "X(top)" is mapped to a button that is physically at the top of the XYAB cluster (being Y), and so on...
- kills many birds with one stone, as it removes the need for creating remap files for the SNES, NES, Game Boy, GBA, and Nintendo DS consoles, and possibly others.

udev/SWITCH CO.,LTD. USB Gamepad .cfg
```diff
@@ -0,0 +1,38 @@
+#http://retro-bit.com/sega-genesis-8-button-arcade-pad-with-usb-m1-black.html
+input_driver = "udev"
+input_device = "SWITCH CO.,LTD. USB Gamepad "
```
miklosaubert, line 3: Where is this value coming from? On my machine (fully up-to-date Arch Linux), `dmesg` shows the following:
```
[12464.638285] usb 1-1: Product: USB Gamepad 
[12464.640034] input: USB Gamepad  as /devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.0/0003:0079:0011.0005/input/input17
[12464.640235] dragonrise 0003:0079:0011.0005: input,hidraw1: USB HID v1.10 Joystick [USB Gamepad ] on usb-0000:00:14.0-1/input0
```
And `lsusb` reports the device like so:
```
Bus 001 Device 008: ID 0079:0011 DragonRise Inc. Gamepad
```
What is the standard here? If the value is arbitrary, why not use the product's real name?

miklosaubert:
I'm the happy owner of a retro-bit 8 button pad, and I was thinking of offering a config file, but I see there is already this PR. I left a couple comments on it, but I also had a general comment:

- Is the file name what we want to ship in the end? Since there's already a config for the 6-button pad named `retro-bit SEGA Genesis 6-Button.cfg`, I think it might make more sense to call this new file `retro-bit SEGA Genesis 8-Button.cfg` or maybe `retro-bit SEGA Genesis 8-button Arcade Pad.cfg`, matching the product name as seen on the retro-bit product page (http://retro-bit.com/sega-genesis-8-button-arcade-pad-with-usb-m1-black.html)

hizzlekizzle:
@miklosaubert historically, we use whatever name RetroArch gives them when you 'save autoconfig profile' in settings > input > port 1 binds.

offalynne:
Filename aside it’s imperative devices matching `ID 0079:0011 DragonRise Inc. Gamepad ` *not* be registered with autoconfig since RetroArch is not able to disambiguate between dozens of devices that use this same board with no reliably detectable identifiers, including flight sticks, arcade joysticks, N64, SNES, Saturn style pads amoung many others.

RobLoach (changes requested):
Looks like some conflicts?

donald-keighley:
Closing this.  Doesn't really make sense since the board is shared between a bunch of different controllers.

PR #605, opened 2019-12-29
Autoconfig for 8BitDo M30 2.4G

sonik-br:
Autoconfig support for 8BitDo M30 2.4G using USB.
Xinput and Dinput


hizzlekizzle:
Hmm, that seems to be reporting itself as a generic xbox/xinput pad, so we won't be able to merge those, as they will conflict with actual xbox/xinput pads

sonik-br:
Ouch. I just figured ou that it uses the same VID and PID. Going to ask 8bitdo if they can change it with a firmware. Closing this PR.

offalynne:
If you want you can follow this format so this config can be used in the future if RetroArch implements further feature detection to help with disambiguation: https://github.com/libretro/retroarch-joypad-autoconfig/blob/93227d68b420a1cb194db04a53d9f30dfface0c3/udev/DIALOG_GP-A11.cfg#L8-L17

emersonnovais:
I have a same problem with this controller on Lubuntu 19.10. My system recognized it as a generic Xbox 360 gamepad and the layout is too confused. I use a DS4 as my default controller and if I change the buttons config, missconfig my DS4 controller. I tryied to modify an input config file but not worked. I used the 2.2 driver version. I bought for mistake the 2.4G version instead the bluetooth version and is a pain in the brain to make it working. Is retroarch have a to save per controller an input config? thanks.

hizzlekizzle:
@emersonnovais that's what the autoconfig profiles are supposed to do, but if 2 pads claim to be the same pad (by reporting the same vid and pid), there's no way to distinguish which is which. You can disable autoconfiguration altogether and then just map your pad in port 1 binds, but you'll have to do it each time you switch pads.

emersonnovais:
They have differents VID/PIDs. I used this number for the 8bitdo from I get by the command lsusb and converted the hex number to decimal but not worked.

hizzlekizzle:
what vid/pid values are you trying?

emersonnovais:
The result of lsusb command is this:

`Bus 002 Device 003: ID 045e:028e Microsoft Corp. Xbox360 Controller`


My sample config is:
[https://pastebin.com/8kvX2TwP](url)

hizzlekizzle:
yeah, those are just generic MS 360/xinput pad IDs. Manufacturers put us in a bad spot when they report the same IDs but have different mappings :(

emersonnovais:
Is there no way to map this controller? My configs are correct?

hizzlekizzle:
Yes, you can map it. You'll just need to disable autoconfiguration and map it in settings > input > port 1 binds. What we can't do is include an autoconfig profile for it, since it reports as an xbox pad but has different mapping.

emersonnovais:
Not worked for me. The Start and select buttons are different when I use the DS4 ou M30, I need to reconfigure these buttons. If I disable the autoconfig option, result in the same problem. May each controller need in separeted ports? Example: DS4 in port 1 and M30 in port 2, for example? Thanks

hizzlekizzle:
You can, but the order can get scrambled each time you boot up your PC (it might, or it might not). If you're using Windows, there's a utility called 'devreorder' that can help with that, though.

PR #606, opened 2020-01-01
iPEGA PG-9083S bluetooth controller

guoyunhe:
I followed the naming of controllers from the same vendor in `android` and `udev` folder. But they are not the same... I didn't find the naming convention guideline. It is good to have. Cases are annoying. I can found both `ipega`, `iPega`, `iPEGA` on the box...


guoyunhe:
can anyone merge this PR? thanks 😿

hizzlekizzle:
These look good, but they conflict with existing iPega configs. If you don't mind commenting out the vid/pid lines, I can merge them in the hopes that we can someday differentiate between pads that share vids and pids.

guoyunhe:
I commented out these lines. Does it mean that this controller will not be recognized correctly until the pid conflict is solved?

hizzlekizzle:
Yes, but if someone wants to use this one instead of the other one, they can just uncomment the lines in this one.

guoyunhe:
I found that in `android/` folder there are many iPEGA gamepads with same pid and vid. Is that okay?

hizzlekizzle:
Yeah, I saw that, as well, and I'm surprised. I thought we stamped all of those duplicates out. I'll have to pick one and comment the rest, I guess :/

guoyunhe:
@hizzlekizzle does RetroArch have the ability to match device names when vid and pid are not unique?

offalynne:
yes, it uses a scoring system that’s based on all three. if “input_device” is correct and unique, it should solve a conflict in case of non-unique vid/pid. make sure to use the “input_name” field to add a more consistent/readable name for the device since the device name is not often adequately descriptive.

PR #610, opened 2020-01-23
Profile for a wired Dualshock 4 V2 controller

ds22x:
An autoconfig profile for a wired (when it shows the blue light) Dualshock 4 V2 controller.

P.S. I know the device name is a bit confusing, but that's the name it provided when it created the profile.


hizzlekizzle:
perfect! thanks!

PR #614, opened 2020-01-26
add DragonRise Inc. Generic USB Joystick for Digger USB 2.0 Joypad from flathub

candrapersada:
input_device is DragonRise Inc. Generic USB Joystick but Joypad is DIGIGEAR JOYPAD USB 2.0


sdl2/DragonRise Inc. Generic USB Joystick.cfg
```diff
@@ -0,0 +1,24 @@
+input_driver = "sdl2"
+input_device = "DragonRise Inc. Generic USB Joystick"
+input_vendor_id = "121"
+input_product_id = "6"
+input_b_btn = "0"
+input_y_btn = "2"
```
RobLoach: The other thing you could add are the labels :wink: 

For example......

```
input_y_btn_label = "4"
```

The RetroPad's Y button is labeled as "4" on the physical gamepad.

candrapersada: yes

PR #618, opened 2020-02-05
Change That fixes RetroLink N64 Controller Compatibility

BrennerLittle:
Switches A and B button to fix comparability.


hizzlekizzle:
Can you explain what this does in a little more detail?

BrennerLittle:
> Can you explain what this does in a little more detail? That is, the motivation and the benefit?

I bought this controller and by default the A and B button were switched. This is a quick fix. I do not know if there is a better solution.

hizzlekizzle:
Would be it be better to change the actual button assignment instead of just the labels?

PR #620, opened 2020-02-07
Added autoconfig for ipega PG-9087S bluetooth in UDEV mode

soredake:


hizzlekizzle:
This conflicts with an existing udev config for an iPega pad.

Is the original one not correct?

soredake:
It ain't worked earlier for some reason, works now.

PR #623, opened 2020-02-20
Add/Update Switch-compatible controllers

nullstalgia:
The [PDP Fight Pad Pro](https://www.amazon.com/Nintendo-Switch-Legend-GameCube-Controller-500-100-NA-D2/dp/B07GX8YS78) is a simple addition, as as such, I don't have much to say about that in particular.

The Pokken controller, though, I have some comments.

[The Pokken controller is often emulated using an Arduino](https://github.com/progmem/Switch-Fightstick) (AVR) to make a custom DirectInput/Nintendo Switch-compatible controller. However, the currently existing profiles aren't suitable for this purpose (missing buttons, little to no labels, not existing at all)

This profile is compatible with both the official controller, and the custom controllers makers have made with that repository I have linked. It would be awesome for our controllers to work natively in RetroArch on any OS if you allow those in.


hizzlekizzle:
These look good, thanks!

PR #624, opened 2020-02-23
Add odroidgo2_joypad.cfg to support odroid go advance

valadaa48:


hizzlekizzle:
Does it have a vid/pid we should include?

natinusala:
Not that I'm aware of, I looked at the kernel driver and dts and couldn't see anything (it's not a USB device).

~~It's "ODROID-GO Advance" by the way~~

PR #627, opened 2020-03-03
Add Nintendo Switch SNES Controller

DanTheMan827:
This adds the SNES Online controller profile when connected over bluetooth


hizzlekizzle:
Thanks for this!

PR #629, opened 2020-03-08
Added Thrustmaster T-Flight Joystick

Jaffacakelover:
Added Thrustmaster T-Flight Hotas X Joystick (PS3/PC) configs


hizzlekizzle:
These look awesome, thanks! Is there a reason there's no vid/pid on HID and SDL2?

Jaffacakelover:
Simple reason, the 'Save autoconfig' option in Retroarch isn't saving any to the cfg file. Should there be? Will it have the same VID/PID across all drivers?

hizzlekizzle:
heh, good reason :)

Yeah, they should be consistent across drivers, though the actual mapping / keycodes can vary. I think adding them would be good, if you don't mind.

hizzlekizzle:
thanks!

PR #630, opened 2020-03-14
Add Retro-bit Saturn Bluetooth controller autoconfigs (Dinput and Xin…

spec-chum:
…put pad modes) for Retroarch Xinput


hizzlekizzle:
the dinput mode one looks good, but it seems the xinput one is just using generic vid and pid, so we can't add it without breaking a lot of other controllers. Would you mind removing it from the PR so I can merge just the dinput one?

spec-chum:
Ah, bloody retro-bit, I should have known,big clue was the device name, in hindsight! 

Didn't realise, sorry. 

Will sort today.

offalynne:
at least the DInput mode's VID/PID are salvagble for this one :cry: :pray:

spec-chum:
Removed :)

hizzlekizzle:
Beautiful, thanks!

PR #633, opened 2020-04-11
Add MayFlash Gamecube Adapter for xinput/udev

flagrama:


flagrama:
Hmm, due to my normal settings I think I have the A and B buttons swapped in these configurations. Should a make a new PR fixing that, or can someone else correct it easier?

hizzlekizzle:
Another PR is fine. Thanks!

PR #636, opened 2020-04-15
Initial version

ReenigneArcher:
Auto config file for streaming RetroArch to moonlight app. Tested with RetroArch on Windows 10 PC and Moonlight on Android TV.


hizzlekizzle:
Hmm, I don't think we can merge this as-is, since it will conflict with the Nvidia Shield controller, which reports the same vid/pid pair. However, if you can comment out the vid and pid, we can merge it in the hopes that we get better at discriminating among conflicting pads in the future. That way we can just uncomment all of the duplicate vid/pids.

ReenigneArcher:
Okay, that explains why it was getting the wrong button mappings. What's the method of commenting in these files?

hizzlekizzle:
just put a pound-sign/hashtag in front of the line you want to comment.

On your local machine, you can comment (or just delete the whole file) nvidia shield controller autoconfig and your custom one will take control.

hizzlekizzle:
whoops I forgot all about this. Thanks for the changes

ReenigneArcher:
Oh, sorry I just closed this to try to make quick change (just naming) as I found out GeForce Experience is controlling the VID and PID of virtual controllers.

Issue #637, opened 2020-04-21
Xbox One Wired config missing for hid (mac osx)


aserrallerios:
> Xbox One Wired Controller (1118/746) not configured.

I'm new here so I don't know if that's a known issue or I'm even doing something wrong.

hizzlekizzle:
That usually just means we don't have an autoconfig profile for it, so you'll need to map it manually, optionally create an autoconfig profile and then optionally submit that autoconfig profile to us for future inclusion :)

aserrallerios:
I saved a working autoconfig, but I'm not able to see the file that contains the config (in order to open the PR here), could you help me?

hizzlekizzle:
if you go to settings > directory, it will tell you where your "autoconfig" directory is located. Your new autoconfig should be in there, not inside any of the subdirectories. While we're at it, once you find it, could you open it in a text editor and add the button labels? You should be able to just copy them from one of the other xbone autoconfigs, like this one:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/Xbox_One_Controller.cfg

Just copy that whole btn_label section at the end of yours and save the file.

aserrallerios:
I'll do that :)

One last thing: what's the approach for mapping the "xbox button"? The one with that looks like the console logo? I've seen no corresponding RetroArch controller correspondence...

RobLoach:
Was merged! Thanks.

PR #638, opened 2020-04-23
Add Xbox One Wired autoconfig

aserrallerios:
As described in https://github.com/libretro/retroarch-joypad-autoconfig/issues/637

It works with the improved OSX driver for the Xbox One controller: https://github.com/360Controller/360Controller#xbox-one-controllers-connected-with-usb


hizzlekizzle:
Looks great. Thanks!

PR #644, opened 2020-05-10
Remove (wired) appendix from profile name

ds22x:
The profile is used by both wired and wireless Dualshock 4 V2's


MochaChocha667:
My idea on Discord.

PR #647, opened 2020-05-14
Make Retro-bit/8Bitdo N64 Controller Work on N64 Cores

tony971:
Since RetroArch N64 cores use retropad "B" and "Y" keys for N64 "A" and "B", the N64 clone controller needs to set "A" and "B" to retropad "B" and "Y" to work.

Note: This leaves the controller without a RetroPad "A" button, so it will break everywhere outside of an N64 core.


hizzlekizzle:
There is a trick that some of the autoconfig profiles use whereby they assign retropad-Y and retropad-A both to the physical B button (I think... you'll want to look around and verify). That lets you keep the menus working while also letting the N64 stuff work. It can be a little weird with other cores, but nothing a remap won't fix.

tony971:
I found one that sets the physical "B" button to both retropad "A" and "Y". If it works as intended, this would make the menu work normally as long as that menu option for inversing accept and cancel buttons on the menu is selected. It would absolutely wreck other cores that actually use both the retropad "A" and "Y" buttons, though.

tony971:
Just tested it. Setting retropad "A" and "Y" to the same button does not work in the menu.

hizzlekizzle:
Here's another way to map that seems to work well for people, if you want to try it: https://cloud.githubusercontent.com/assets/10035308/16599635/7f35579a-42c0-11e6-9615-9a3670932332.png

RobLoach (changes requested):
Looks like some conflicts?

RobLoach:
Going to close this. Been a while, and there's some conflicts.

PR #649, opened 2020-06-06
Add support for Xin-Mo Dual Arcade config based on

Ntemis:
Playstation and Playstation 2 key legend
provided by X-arcade manual here https://ibb.co/yVfsdyb


hizzlekizzle:
Hmm, I would normally ask for button labels, but I guess xarcade doesn't have any labels normally, eh? I'll go ahead and merge it and if you come up with any labels that make sense, you can just do another PR :)

PR #652, opened 2020-06-08
Add d-pad up to the Google Nexus Gamepad config

manojka:
D-pad up is missing when I use my Nexus Gamepad, and this looks like why. It's untested, as I don't know how to test a gamepad config in unrooted Android without getting a whole build environment up, but this change looks like it can't cause harm, and the current config is definitely broken.


hizzlekizzle:
lol yeah, good catch :)

PR #653, opened 2020-06-09
Fix up more Google Nexus Controller button bindings

manojka:
I figured out how to edit controller autoconfig without root, so here I think are most of the changes needed to make this functional. Unless specified, I tested the button changes in a game.

- Fix typo to make X and Y buttons work correctly
- There's no select button, so bind it to the back button like on the Nvidia Shield controller
- Change R2 trigger (done through UI's binding system, untested because I don't have a game loaded that uses it yet, but it also matches the Shield controller's binding)
- add and rename some labels for clarity and correctness

There's one remaining problem that the provided start button code is the default KEYCODE_BUTTON_START, but this gamepad has no start button. The middle button should have a keycode that works, but it's being flaky for me, so I've left that alone for now.


hizzlekizzle:
This looks good! Thanks for working on it.

manojka:
Glad to do it, and thanks to you, too!

To finish the job I'm trying to bind the power button short press, keycode 110 (KEYCODE_BUTTON_MODE) to "Start". The Nvidia Shield OS is detecting the button press every single time, and RetroArch detects it every time in the binding interface (though I'm setting it manually because holding the button down powers the controller off before RetroArch accepts the new binding). But it gets detected only unreliably in the game itself. I don't have that problem with other controllers, so it seems like a combination of that particular controller, button code, and RetroArch, and in no other cases (so far).

Any chance you'd have an idea how to debug this further? Thanks.

manojka:
Long story short, I changed my testing apparatus, and if I believe the "Gamepad Tester" app on another device, that middle button I want to make work is behaving weirdly in the OS, so not RetroArch's fault.

I think I'll submit another PR later to make the R3 thumbstick press act as start, unless you or someone suggests otherwise.

PR #657, opened 2020-06-14
Give Google Nexus Controller a start button

manojka:
There's no start button on this controller, so I took what I guessed would be the least used button, R3, and bound it to the start button. I'm don't know a better choice, but I'm open to ideas.

I couldn't bind the power button (which produces KEYCODE_BUTTON_MODE, 110, when short-pressed) to the start button because RetroArch on Android detects it unreliably, I think because of how the OS is supplying the keycode (which might be due to the OS or controller itself, I'm not sure). Duplicated on two different Android test devices.

I also made the labels for the left and right shoulder buttons consistent with the others in the file.


hizzlekizzle:
Since that's super-weird (though I understand and agree with your reasoning), would you mind adding a comment in the file explaining that you mapped start to R3 due to the lack of a true start button? That way anyone who is confused and comes across it will understand what's happening and why.

manojka:
Ah, I didn't know these files even supported comments. I guessed how they were formatted and that worked. I think I did the right thing to make the comments show up in this pull request.

hizzlekizzle:
Looks good, thanks!

PR #667, opened 2020-07-11
Add Quasicon 2P Arcade Control (udev)

mrestifo:
Added autoconfig file for Quasicon 2P Arcade Controller (udev)

- default setup for all buttons
- labels for buttons and axis

afdea5c426de41ac8d2faeca88d6191b


hizzlekizzle:
that's a wild controller :O

Autoconfig looks good, thanks!

PR #670, opened 2020-07-17
Fixed 8BitDo M30 udev mappings for firmware 1.14

Sanaki:
Xinput over USB for 1.13 and 1.14 isn't viable unless we can somehow differentiate between devices reporting as an xbox 360 controller.
Xinput over USB can be used if the firmware is downgraded to 1.10.
Given the lack of rumble motors, dinput is recommended for this controller.


Sanaki:
To be clear, `8BitDo M30 gamepad.cfg` was converted to `8BitDo_M30_BT_B.cfg`, but is seen as a different file due to the extent of the update.

PR #671, opened 2020-07-18
Update Arcade Joystick v1.1.cfg

lfanog:
Added "display name" / Changed "vendor id" and "product id"

### DIFF
--- a/xinput/Arcade Joystick v1.1.cfg
+++ b/xinput/Arcade Joystick v1.1.cfg
@@ -1,8 +1,9 @@
 input_driver = "xinput"
 input_device = "Arcade Joystick v1.1"
+input_device_display_name = "Arcade Joystick"
 
-input_vendor_id = "33410"
-input_product_id = "32769"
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
What's up with the different vid/pid? is the other one incorrect?

lfanog:
> 
> 
> What's up with the different vid/pid? is the other one incorrect?

I didn't observe it before. I checked now on USBDeview.

lfanog:
Arcade Joystick v1 1

hizzlekizzle:
I believe the old one is correct and the new one is the hex value. That is, 8282 is hex for 33410.

lfanog:
> 
> 
> I believe the old one is correct and the new one is the hex value. That is, 8282 is hex for 33410.

Yes, you are correct.

:)

twinaphex:
@hizzlekizzle So what do we do with this PR?

lfanog:
I will close.

PR #682, opened 2020-09-04
Fix various mislabels and adds MFI driver autoconfig functionality

WeedyWeedSmoker:
This fixes various mislabels, adds correct labels to the MFI/MFI-compatible controller buttons, and adds a proper configuration file for the MFi driver…

**(The MFi specification now supports DUALSHOCK®4 and Xbox Wireless Controller as well as the original MFi-certified controllers whose buttons are labeled like the Xbox ones, hence the requirement for double-labeling like "A/Cross" and "X/Square"…)**

@twinaphex @hizzlekizzle


WeedyWeedSmoker:
@twinaphex Hi!

Do you think this should be merged as-is or should I change back "`hid/Nimbus.cfg`" to use L2/R2 as select/start? (As it only has 1 button in the center, which is labeled MENU…)

Thanks!

WeedyWeedSmoker:
@twinaphex You can merge this, the mislabelled buttons had to be fixed, and the MFI driver is tested correctly…

WeedyWeedSmoker:
@hizzlekizzle Do you want to merge this? I don't think there's any problem with this commit as far as I can see…

Thanks!

PR #683, opened 2020-09-04
HID: add HORI mini wired PS4 USB controller

Kert:
Related to https://github.com/libretro/RetroArch/pull/11280


RobLoach (changes requested):
Thanks a lot! Would you mind adding a comment at the top of the config with a link to the hardware?

Kert:
@RobLoach Would this be fine? https://stores.horiusa.com/wired-mini-gamepad-black/. Does // work there?
Also the other configs don't have any links included

RobLoach:
That URL is perfect. You'll want a comment about the device, along with a `input_device_display_name` so that the notification looks pretty when you plug it in.......

Example: https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/3H%20Dual%20Arcade.cfg

hid/HORI_mini_wired_ps4.cfg
```diff
@@ -0,0 +1,55 @@
+input_device = "HORI mini wired PS4"
```
RobLoach, line 1:
``` suggestion
# HORI Wired Mini Gamepad for PS4
# https://stores.horiusa.com/wired-mini-gamepad-black/
input_device = "HORI mini wired PS4"
input_device_display_name = "HORI PS4 Mini Wired Gamepad"
```

Issue #642, opened 2020-05-07
DualShock 4 autoconfig problem on android


klepp0906:
Fortunately I have two Nvidia shields so was able to track this down with relative ease. 

Connected my ds4 via usb to one and fully updated.  Suddenly I couldn’t enter the menu from games which I had bound to start+select. 

I tracked it down to being the “select” half of the equation. 

After testing on the other shield which I could enter the menu I updated everything BUT the autoconfig.  Could still enter the menu. 

Updated autoconfig, poof - issue. Fortunately I backed up the older one so I’m not screwed,  but figured I’d better mention it so I don’t have to refrain from updating AND remember to hold the old autoconfigs perpetually lol.

klepp0906:
K - just dove into the files and found it pretty easily

`Sony_Dualshock_4_Controller_USB.cfg` in the android folder has `input_select_btn = "4"` in the working version and `input_select_btn = "109"` in the non working version.

no idea why the change, other than the fact that you would "expect" the select button to be 108 considering start is 109.  Via testing that simply isnt the case.

mapped it manually and checked the resulting cfg myself. 

manually changed via a text editor on the other shield and it began working again etc.

RobLoach:
A few changes were merged forwards. Thanks!

PR #686, opened 2020-09-08
Fix Android DS4 config select bindings and rename configs

andshrew:
Reference issue: https://github.com/libretro/retroarch-joypad-autoconfig/issues/642

The current configs for the DualShock 4 have the Touchpad configured as the select button, rather than the more conventional select button which would be the Share button.

This PR reconfigures the Share button to be the select button, and the Touchpad to be the RetroArch menu button (as far as I can tell the PlayStation button is unusable on Android as the system binds it as the Android home button).

Additionally this renames the configs for consistency; there should be two configs as there are two revisions of DS4.

This has been tested on an Nvidia Shield TV with v1 and v2 DS4s both connected by cable and bluetooth.

One inconsistency remains between the configs in that the v2 config has `input_r2` and `input_l2` bound to an axis instead of a button. Rebinding in application results in it defaulting to a button configuration, but as R2 and L2 seem to work fine in this axis configuration I've left this unchanged given that it seems to work fine.


bslenul:
That's a weird one, maybe it varies between Android version or something? My phone is still under 7.0 and when I had my V1 Share button was "109".

Now I have a V2 and my bindings are completely different, this is what I have when rebinding:
```
input_device = "Wireless Controller"
input_vendor_id = "1356"
input_product_id = "2508"
input_b_btn = "97"
input_y_btn = "96"
input_select_btn = "104"
input_start_btn = "105"
input_up_btn = "h0up"
input_down_btn = "h0down"
input_left_btn = "h0left"
input_right_btn = "h0right"
input_a_btn = "98"
input_x_btn = "99"
input_l_btn = "100"
input_r_btn = "101"
input_l2_btn = "102"
input_r2_btn = "103"
input_l3_btn = "109"
input_r3_btn = "108"
input_l_x_plus_axis = "+0"
input_l_x_minus_axis = "-0"
input_l_y_plus_axis = "+1"
input_l_y_minus_axis = "-1"
input_r_x_plus_axis = "+2"
input_r_x_minus_axis = "-2"
input_r_y_plus_axis = "+3"
input_r_y_minus_axis = "-3"
input_menu_toggle_btn = "110"
```
\+ touchpad being "106" (and yes, "110" is my PS button and it works fine for me when bound as "toggle menu").

andshrew:
It's interesting that your controller is identified as `input_device = "Wireless Controller"` rather than `Sony Computer Entertainment Wireless Controller` which I guess suggests a different driver is being used.

It looks like Android 9 added native support for DS4 which could explain the difference you see.

I've also tested on an Android 7 phone now - albeit a Sony Xperia (so it had official support for DS4) - and while that also reports as just `Wireless Controller` the actual bindings match this config although in this case neither the touchpad or home button work.

It seems like Android may just too many variations to fit the potential bindings for the DS4 into the current RetroArch controller map selection system. It could possibly do with some additional selection criteria based on OS version.

I'd be inclined to suggest that bindings based on the official DS4 implementation from Android 9 should be used for the default auto config.

bslenul:
Funny (and annoying...) how it can vary from a device to another :p 

> I'd be inclined to suggest that bindings based on the official DS4 implementation from Android 9 should be used for the default auto config.

Yeah, that would make sense IMO 👍

offalynne:
Looks like it's mapped differently on Android 9 forward.

RobLoach:
Merging this one. Let me know if there are any other updates needed.

PR #687, opened 2020-09-09
add config for Xbox 360 RF module Rev VA

bhartshorn:
This is for an Xbox 360 RF module converted to USB. The actual config is directly copied from `Microsoft X-Box 360 pad.cfg` Tested and works as intended.

Is there a way to specify multiple `{vendor,product}_id` pairs in a single file? There are lots of revisions of these things that probably all have different product_id but function identically.


offalynne:
> Is there a way to specify multiple {vendor,product}_id pairs in a single file?

No, you have to duplicate and edit the fields per file.

bhartshorn:
> Presumably autoconfig can pick them up based on device name match?

I'm not familiar with how these are matched (other than USB ids, obviously) - do you know? 

I can tell you that the adapter, or rather attached controllers did not get picked up by one of the existing config files, hence creating a new one. :+1:

offalynne:
> I'm not familiar with how these are matched

they're scored based on the fields `input_device` `input_vendor_id` and `input_product_id`, so you can omit whichever varies (presumably product id) and as long as there aren't any conflicting configs it should be get picked up correctly

bhartshorn:
I see. Thank makes sense. This device is actually recognized properly by the existing autoconfigs in retroarch installed on Archlinux, but not in Lakka on my raspberrypi (should be latest version). 

I didn't check all the other configs when I created this - it might be that Lakka doesn't have the latest set or something. Regardless, this PR may not be necessary. Not sure if I'll have time to dig deeper right now.

PR #688, opened 2020-09-11
(XINPUT) XBOX 360 and Generic XInput touch-up

sonninnos:
- ~~Added proper official XBOX 360 profile to prevent fallback usage~~
- Added proper official wireless XBOX 360 profile to prevent fallback usage
- Removed obsolete "(User X)" profiles
- ~~Renamed "Generic 360 Pad" to "Generic XInput Controller"~~
- Renamed "Generic 360 Pad" to "XBOX 360 Wired Controller"
- Reorganized btns and labels

~~With these changes DualShock3 via ScpToolkit will be seen as "Generic XInput Controller" and regular X360 controllers as "XBOX 360 Controller", matching the XBOX One profile display name.~~
With these changes DualShock3 via ScpToolkit and wired controllers will be seen as "XBOX 360 Wired Controller" and regular wireless X360 controllers as "XBOX 360 Wireless Controller", matching the XBOX One profile display name.


andshrew:
Is your product ID (673) for the 360 controller definitely correct? My official wired controller has product ID 654 which you have in the `Generic XInput Controller` config here. 654 is also the ID which is already used in the dinput config (XBOX_360.cfg)[https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/dinput/XBOX_360.cfg]. 

If the user 2,3,4 configs are no longer required it may be better to just have a single config for this rather than trying to differentiate between official and generic 360 controller, especially given the number of controllers which use these IDs to appear as 360 controllers.

sonninnos:
Ah, that explains why ScpToolkit controller is seen as different product_id. I only have actual wireless ones to test with.

So perhaps the generic should be renamed as "Wired" and this new one "Wireless" then.

Single config indeed won't cut it, since the reason for these changes was because the wireless ones went into fallback mode due to the recent changes in RA that require product/vendor matches.

Yes, those extra 2,3,4 configs did nothing. The correct new config was picked up fine for both 2 identical wireless ones, but there is an unrelated issue in RA when mixing wired and wireless: If wireless is #1 and wired #2, the controller labels will switch around, so that the first controller will be seen as the second (both in the init widget and in remap index), but they do work in the correct order.

andshrew:
> So perhaps the generic should be renamed as "Wired" and this new one "Wireless" then.

I think that sounds like a good idea.

sonninnos:
Hopefully that sorts it out.

xinput/XBOX 360 Wired Controller.cfg
```diff
@@ -1,18 +1,17 @@
 input_driver = "xinput"
 input_device = "XInput Controller"
-input_device_display_name = "Generic 360 Pad"
-
+input_device_display_name = "XBOX 360 Wired Controller"
```
RobLoach, line 3:
``` suggestion
input_device_display_name = "Xbox 360 Wired Controller"
```

xinput/XBOX 360 Wireless Controller.cfg
```diff
@@ -1,14 +1,17 @@
 input_driver = "xinput"
-input_device = "XInput Controller (User 1)"
+input_device = "Controller (XBOX 360 For Windows)"
+input_device_display_name = "XBOX 360 Wireless Controller"
```
RobLoach, line 3:
``` suggestion
input_device_display_name = "Xbox 360 Wireless Controller"
```

RobLoach (changes requested):
Looks good. Thanks! Just a quick update to use `Xbox` instead of `XBOX`.

sonninnos:
Okey dokey!

Issue #685, opened 2020-09-06
RetroArch 1.9.0 breaks all Switch controller autoconfigs on Xinput


nfp0:
The new release of RetroArch 1.9.0 changes the way the autoconfig lookup works for Xinput and Dinput devices.
According to @hunterk :
> with 1.8.9, xinput and dinput looked for profiles within each others' groups, which led to unpredictable behavior. We stopped that in 1.9.0, but it's led to some controllers that were previously autoconfigured to no longer be so
The solution is: you can either map your controller(s) and make your own profile(s) and then preferably submit it/them to the repo for inclusion, or, if you can identify the one it used to use, just make a copy of it and move it to the folder for the driver you use and change the "driver" line in the file to match

I can confirm this affects the very popular Switch Pro and Switch SNES official Nintendo controllers. They are no longer auto-configured in RetroArch on Xinput (which is the default driver on Windows). Using the workaround described by @hunterk works, but it would be nice if this was fixed in this repository.

Note: This also affects the Steam integration autoconfigs for the Switch Pro controller.

Also, I haven't tested if the Joycons are affected, but I would assume so.

bslenul:
If you and/or someone else can confirm these dinput autoconfigs work: [switch_dinput.zip](https://github.com/libretro/retroarch-joypad-autoconfig/files/5225319/switch_dinput.zip) (extract in retroarch/autoconfig/dinput/) I'll make a PR ;)

nfp0:
I tested those and I can confirm they work (with the exception of the NES controller, because I don't have one).
But the Switch Pro Steam integration is still not working. I uncommented the `input_vendor_id` and `input_product_id` lines from [this file](https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/Nintendo%20Switch%20Pro%20Controller%20USB%20Steam%20User%201.cfg) and it started working again. But I see those lines were commented for a reason, so I don't know if this is a good enough solution.

bslenul:
Thanks for confirming 👍 I won't touch the Steam thing but PR done for the other ones (NES included, if it works for the other 3 I don't see any reason why it wouldn't work for this one) :)

RobLoach:
Looks like these were merged? Feel free to open a new issue if you have any other problems.

PR #691, opened 2020-09-16
Copy Switch Pro, Joy-Con and S/NES controllers to dinput folder.

bslenul:
For #685.


bslenul:
Files were included in #701, so this is useless now, closing (and the related issue can be closed as well)!

PR #692, opened 2020-09-20
Disable Profiles Causing CI Travis Errors

andshrew:
The CI check for duplicate controller profiles has been failing for some time. This PR disables the configs which are causing the failures.

`xinput/Nintendo Switch Pro Controller USB Steam User 1.cfg`
This was already disabled due to a vid/pid conflict with `Steam_Link_Virtual_Pad.cfg` but was still reporting a conflict as input_device also needed to be commented out for the CI script to detect it as disabled.

`udev/SWITCH CO.,LTD. USB Gamepad .cfg`
The vid/pid conflicts with an older existing profile `retro-bit SEGA Genesis 6-Button.cfg`. It looks like Retro-Bit have used the same pid for two different controllers as the bindings differ. Disabling this profile only on the basis that this one was submitted after the `retro-bit SEGA Genesis 6-Button.cfg`.


offalynne:
@hizzlekizzle @twinaphex

hizzlekizzle:
hey sorry for sleeping on this. Looks like the Switch Pro / Steam configs have merge conflicts. Can you get those fixed up?

andshrew:
No problem, I've merged the two and only `udev/SWITCH CO.,LTD. USB Gamepad .cfg` needs to be changed now. Just waiting for the CI check to verify it but it ran ok when run locally.

RobLoach (approved):
While this approach is not optimal, I think it is a good workaround. When the vendor/product matches with the other config, it'll choose that instead. Thanks!

PR #697, opened 2020-10-02
Add Qanba Q4RAF arcade stick

bestie:
The mapping I got for this controller was incorrect so here is the mapping I created and have been playing with for a few months.

I believe the `ACRUX_QuanBa_Arcade_JoyStick_1008.cfg` is incorrect in calling device `3888:4352` a Q4RAF, I think it's actually a Q1-W 2-in-1 which is quite different.

I can remove the comment or change the name in other config as required.

### DIFF
--- a/udev/QanBa_Q4RAF_Arcade_Stick.cfg
+++ b/udev/QanBa_Q4RAF_Arcade_Stick.cfg
@@ -1,0 +1,72 @@
+# Jess Technology Co., Ltd
+# QanBa Q4RAF Arcade Stick
+# Decimal vid:pid = 3888:4111
+# Hex vid:pid = 0f30:1012
+
+# PS3/PC mode
+
+# | Q4 Label | Button  | Retropad |
+# | A        | 0       | B        |
+# | B        | 1       | A        |
+# | X        | 2       | Y        |
+# | Y        | 3       | X        |
+# | R1       | 4       | L1       |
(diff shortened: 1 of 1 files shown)


offalynne:
Is the `input_name` field in your config edited, or is it what the device sets by default?

bestie:
The whole file is edited. I just copied it from `udev/ACRUX_QuanBa_Arcade_JoyStick_1008.cfg` which was the config the controller was initially matching up with.

Are you referring to this line?
`input_device = "QanBa Q4RAF Arcade Stick"`
Is there an authoritative source to get that string from? 

BTW I removed the comment from `udev/ACRUX_QuanBa_Arcade_JoyStick_1008.cfg`, I guess that's a separate issue.

offalynne:
“input_device” is a description provided by the device. remove matching configs and see what’s assigned, or check with js-test or lsusb.

“input_device_display_name” is for manually assigning a name.

bestie:
Thanks for pointing me in the right direction. I re-read the guide (I initially missed the device name part) and took a closer look at some of the other mappings.

Now there is:
- Full `input_device` name  from `lsusb`
- Friendly name that will make sense to users of this device
- Removed non-functional home button entry
- Hopefully made this relatively consistent with other configs

Apologies for the initial confusion, I really appreciate your work in this project!

PR #700, opened 2020-10-18
Add profile for X360 controller with wireless receiver.

bslenul:


bslenul:
Oh my bad, there's already a PR (#688) which handles it, I'm closing this!

PR #701, opened 2020-10-20
Xinput/dinput cleanup.

bslenul:
Moved every non-xinput profiles (VID != 1118) from xinput to dinput folder and changed `input_driver = "xinput"` to `input_driver = "dinput"` for these profiles.

Since 1.9.0 RetroArch ONLY checks in dinput folder for dinput controllers even if xinput is selected as driver, it should prevent all these issues people are having because their old profiles are still in xinput folder instead of dinput.

I was making little PRs previously by moving xinput profiles to dinput folder each time I caught someone with the issue, but with that logic we will end up with twice the amount of files, which doesn't seem like a good idea... I was doing it for backward compatibility reason but like Elmo said on Discord:

> Is it likely that people running old builds are updating autoconf?

Which seemed like a good point to me :D 

Anyway, @hizzlekizzle @jdgleaver seems OK for you guys? Jdgleaver I'm pinging you as well because you made the xinput/dinput changes, so you'd probably be the best person to catch an issue with this logic :p


jdgleaver:
This looks great to me - I can't find any fault with the logic :)

Good work!

PR #703, opened 2020-11-08
Added Austgame PS2 to USB adapter

rshamsnejad:


rshamsnejad:
The checks indicate that there are two duplicates, which is not true because they have different product and vendor id.

RobLoach (approved):
Thanks!

PR #706, opened 2020-11-14
add DaemonBite Arduino board linux configs

hunterk:


RobLoach (approved):
This is great! Thanks for adding to add the input devices too.

PR #707, opened 2020-11-15
Create Saturn to USB v3.0.cfg dinput file for Raphnet Tech Saturn to USB adapter v2

roludoM:
Hello, this is my first contribution and hope I did well and that it will help.

This is an autoconfig dinput file for the latest Rapnet Technologies Saturn to USB adapter v2. 
https://www.raphnet-tech.com/products/saturn_to_usb_adapter_v2/index.php

This works with the standard saturn control pad and also the saturn 3D control pad (both modes standard / analog with analog stick & analog triggers working with saturn beetle core)

Depending of the mode selected on the 3D control pad, D-pad & Triggers don't have the same mapping. Still every case is covered by this config file. That is why some inputs for the D-pad & the Triggers looks to be duplicates, but in reality it works with no conflicts.


RobLoach (approved):
Thanks! This looks great. Appreciate your adding documentation comments at the top.

PR #709, opened 2020-12-12
3 dinput profiles had "winxinput" as "input_driver".

bslenul:
I don't even think it breaks anything by leaving these to "winxinput", but let's make it clean! :)


RobLoach (approved):
Good clean up, thanks.

PR #714, opened 2020-12-25
add udev HORI Fighting Stick Mini (Nintendo Switch)

whatdoineed2do:
The HORI stick is misidentified and picked up as either the `Gamester Reflex Arcard stick` and `Generic X-Box pad` profiles.

Dedicated profile


RobLoach (approved):
Looks pretty good. Thanks a lot!

PR #715, opened 2020-12-30
add Xbox Series X/S mappings

alkazar:
Adds support for Xbox Series X/S controllers.

Tested with both the official Microsoft wireless dongle (using the xow driver) and a wired USB connection.

Bluetooth is currently non-functional on Linux with this controller.


RobLoach (approved):
Looks good! Thanks a lot.

PR #717, opened 2021-01-07
Add Go Advance gamepads for all 3 generations

valadaa48:
Sorry, HK renamed all of their devices.

We left the old ones in there so we don't break units that don't have the updated kernel yet.


RobLoach (changes requested):
Thanks! Is there `input_vendor_id` and `input_product_id` info for each???

valadaa48:
Not sure if this will work since 2 of them have dup product id's and only differ by version. I didn't see a input_version_id to help differentiate:

```    [ model name, vendor, product, revision ]
    "GO-Advance Gamepad", 0x484B, 0x1000, 0x0100
    "GO-Advance Gamepad (rev 1.1)", 0x484B, 0x1000, 0x0101
    "GO-Super Gamepad", 0x484B, 0x1100, 0x0100
    "GO-Next Gamepad", 0x484B, 0x1200, 0x0100
```

PR #719, opened 2021-01-14
3 new autoconfigs (sdl2) + 1 fix (udev)

mrmatteastwood:
Logitech F310 Gamepad (XInput), Afterglow, Generic X-Box pad


mrmatteastwood:
Hey guys, for the Afterglow udev one, I noticed that in the repo, there's a file called,

Afterglow_Gamepad_for_Xbox_360.cfg

That file gets loaded when plugging the controller into RA. When I re-configure the controls though and save a new file, its file name is:

Afterglow Gamepad for Xbox 360.cfg

When resetting to default afterwards, the faulty mappings from the first file are loaded again and the newly created file is ignored.

Looks like you'll need to delete the one with the underscores in the file name...

RobLoach (changes requested):
These look great! Thanks a lot. Would you mind adding a comment at the top of each config with a link to the hardware for documentation? Thanks!

mrmatteastwood:
Hey Rob, no problem. :-) I made the changes. Can you see them now, or is there something I need to do to this pull request / do I need to open a new one?

Also, please bear in mind that in the udev folder, the following file needs to be deleted:
Afterglow_Gamepad_for_Xbox_360.cfg

It contains the incorrect mapping for some buttons. When that file is present, RetroArch loads it instead of the correct one I created (with RetroArch).

mrmatteastwood:
Okay, looks like I needed to create a new one. It's #729.

PR #725, opened 2021-01-24
Update 8Bitdo_N30_Modkit_BT.cfg

offalynne:


RobLoach:
This is most excellent. Thank you!

Sanaki:
Unfortunately this profile is conflicting with the Xbox One profile. Short of disabling it by commenting out the vid:pid, is there something I'm missing that would allow fixing this issue? I've seen multiple people on discord complaining about it now.

PR #735, opened 2021-02-17
Adds DualShock 4 v2 mapping for sdl2

soredake:


soredake:
Same as "PS4 Controller.cfg" but different `input_product_id`

PR #743, opened 2021-03-16
Added 8BitDo Pro 2 gamepad

ZackXu233:
The hid configuration is generated on my Mac computer, I am not sure whether to use'hid' or'iohidmanager', at least it works on my MacBook Pro.


offalynne:
Sweet! Do the back buttons (P1/P2) generate unique button events if they’re not mapped with the Ultimate Software?

Thanks!

ZackXu233:
No, it defaults to null. There will be buttons only after you configure it yourself.

PR #756, opened 2021-04-20
Update sdl2 config for Nintendo Switch Pro Controller

sbogomolov:
With the current config D-Pad keys do not work. Mappings in this update
are mirroring mappings in the updated udev config for this controller.
These mappings are tested with the new hid-nintendo driver, which
supports both wireless and USB connections of the controller. This
driver was supposed to be included in the kernel from 5.10, but it was
not. The easiest way to get it is to install dkms-hid-nintendo
(https://github.com/nicman23/dkms-hid-nintendo).


offalynne:
perhaps best to wait till hid-nintendo actually ships?

sbogomolov:
This is a valid point. Here are some things to consider:
- Currently this driver is the best way to use Nintendo Switch controllers: it supports both wireless and USB.
- Similar change was already merged to udev branch, so it might make sense to keep things in sync.
- Looking at what was changed in the udev commit (https://github.com/libretro/retroarch-joypad-autoconfig/pull/732), I'm not sure that current mappings actually work even without hid-nintendo driver.

It might be a good idea to keep old mappings with commented out pid and vid lines (similar to what was done for udev update).

RobLoach:
Looks like some changes went through, resulting in a conflict. My bad! Mind a clean up a little bit?

sbogomolov:
@RobLoach Done, should be good to merge now. I have also added the old version of the mappings (disabled by default) in case this has ever worked for someone.

RobLoach:
Thanks a lot for the update.

PR #757, opened 2021-04-29
Fix various configs with btn-only labels for axis triggers

Sanaki:
This fixes various configs showing the axis ID instead of the label for the triggers.


RobLoach:
Thanks! Good find on those.

PR #758, opened 2021-05-02
Add Atomic PSX/PS2 Pad Converter for Linux

davi92:


sdl2/GAMESAIKE USB GAME Controller .cfg
```diff
@@ -0,0 +1,55 @@
+input_driver = "sdl2"
+input_device = "GAMESAIKE USB GAME Controller "
+input_vendor_id = "1241"
+input_product_id = "3862"
+input_device_display_name = "PlayStation Controller"
```
RobLoach, line 5: Could we change the display name for htese? They're not really playstation controllers.

RobLoach (changes requested):
Looks great! Could use an update on the display name though.

davi92:
The device is just a converter for psx/ps2 joypads only, so just like the profile for xinput, I gave it the same name.

PR #763, opened 2021-06-02
xinput: Add MAYFLASH UAS F500 V2 support

swmarc:
Add support for the MAYFLASH Universal Arcade Stick F500 V2 in XInput mode


xinput/MAYFLASH Arcade Fightstick F500 V2.cfg
```diff
@@ -0,0 +1,48 @@
+input_driver = "xinput"
+input_device = "MAYFLASH Arcade Fightstick F500 V2"
```
RobLoach, line 2: Could you provide a `input_device_display_name`? Thanks!

swmarc: Thanks for the hint! Looks way better in RetroArch.
Just in case, that the arcade stick is labelled differently than what it reports technically.

RobLoach:
Looks great! Thanks a lot.

PR #764, opened 2021-06-07
Add SN30 Pro for Android gamepad

8BitDo:
The USB and BT information of this product are the same.


android/8BitDo_SN30_Pro_for_Android.cfg
```diff
@@ -0,0 +1,64 @@
...
+input_start_btn_label = "Menu"
+input_a_btn_label = "B"
+input_x_btn_label = "Y"
+input_l_btn_label = "LB"
+input_r_btn_label = "RB"
+input_l2_btn_label = "LT"
+input_r2_btn_label = "RT"
+input_l3_btn_label = "LSB"
+input_r3_btn_label = "RSB"
+input_menu_toggle_btn_label = "Xbox"
```
hizzlekizzle, line 50: Why did you decide to label this button "xbox"? I didn't see it labeled as such in any images of the controller I saw online, though I'll admit that I'm not sure I was looking at the right one(s).

8BitDo: This product is licensed by Microsoft, so the button name is the same as the Xbox controller. Maybe you can call it "Guide" Button like the previous configuration?

hizzlekizzle: it's purely for end-user identification purposes, so whatever makes it easiest for them. We typically go with whatever label is physically present (e.g., if it's a house, we call it 'home', if it's a logo, we call it 'guide', etc.). It's not set in stone or anything.

8BitDo: I understand what you mean, it is a button with the Xbox logo on it, you can check the pictures on our [website](https://www.8bitdo.com/sn30-pro-for-xbox-cloud-gaming-on-android).

hizzlekizzle: Oh, I see. Yeah, I was looking at completely different pads. "xbox" is perfectly appropriate :)

PR #771, opened 2021-07-03
Update and rename Thrustmaster eSwap Pro PS3.cfg to Thrustmaster eSwa…

Sean-Paulk:
…p Pro.cfg

Added spacing and removed the "PS3" because the controller is for the PS4.


RobLoach:
Thanks!

PR #772, opened 2021-07-08
Update Amazon Fire TV Remote.cfg

fpscan:
Current config is back button enters and enter button goes back.
FireTV Stick 4K


fpscan:
@RobLoach

PR #775, opened 2021-07-15
Create BrawlerGen USB Genesis Mini (Switch mode).cfg

Hew-ux:
New Retro Fighters controller. https://retrofighters.com/our-collection/brawlergen-usb-genesis-mini-switch-mac-pc-controller/

When connected to a USB 2.0 or lower port, it will automatically go into Switch mode. The button labels have been written with respect to what is physically written on the controller despite it using the Switch Pro Controller inputs, making this transparent to the user. Supports the macros as well as the virtual home button (Start+A).


dinput/BrawlerGen USB Genesis Mini (Switch mode).cfg
```diff
@@ -0,0 +1,56 @@
+input_driver = "dinput"
+input_device = "Pro Controller"
+input_device_display_name = "BrawlerGen USB Genesis Mini"
+input_vendor_id = "3853"
+input_product_id = "193"
+
+
```
RobLoach, line 7:
``` suggestion
```

PR #781, opened 2021-07-27
Update Microsoft_X-Box_One_pad_(Firmware_2015).cfg

lbschenkel:
This mapping is missing the menu toggle button and all labels.
Missing entries were copied from `Microsoft_X-Box_One_pad.cfg`.

I personally tested this mapping in my system with my controller.


lbschenkel:
Amended commit message. Original message mentioned that my USB ids did not match the profile, but I haven't noticed that the file uses base decimal and not hex. The USB ids in the profile match the ones in my system.

PR #789, opened 2021-08-31
Updated blissbox firmware 3.X config files

SteveFox1620:
This is the current collaboration of config files I have been hosting for blissbox. These files only work with the newer 3.x firmware files and fix all mappings and have been tested thoroughly with multiple people. Everything should be in proper formats and include information in the headers when needed. If any changes need to be made to conform better to RA formats please let me know. All driver folders are included AFAIK. Thank you.


RobLoach (changes requested):
Looks like a lot of unrelated changes.

skunx99:
Steve is maintaining the mappings for Bliss-Box's firmware 3.0 RA was originally designed for firmware 2.0. Not very many people use 2.0 anymore so he is requesting to update the mapping as they changed on the major release. 

FYI: Bparker wrote an API to get info from the Bliss-Box, that same API could be used to get the version 2 or 3 but code would need to be written for this. This could be utilized to created names like "Bliss-Box 4-Play **3.0** ATARI Port 4.cfg" allowing RA to support both firmwares. If that is desired please reach out for help. Otherwise it should be commonly understood that 3.0 will be the greater of the two used.

RobLoach:
Going to close this one for now, as there aren't been any movement. Feel free to re-open a new Pull REquest with some focused updates. Thanks!

SteveFox1620:
The only thing to move is the push. Thanks. =)

PR #790, opened 2021-09-01
fix Mayflash SS to USB Adapter

barbudreadmon:
previous config doesn't work and doesn't make sense either, replacing it by the one i've been using for years


bslenul:
Autoconfigs for these controllers should probably match the default mappings of the cores, no? Like this I mean:

 layout1-3

https://github.com/libretro/RetroArch/issues/12849#issuecomment-903779082

Idk if there's an "official guideline" for these cases. You also have the "B" label twice (one should be "Y").

barbudreadmon:
The saturn cores aren't mapped like this though ? 
I don't think it's a good idea to reinvent the wheel : those SS controllers are fightpad, so they should probably be mapped like standard fightpad : 
fusion1
madcatz
hori
razer
fusion

bslenul:
> The saturn cores aren't mapped like this though ?

They are, the only exception being Beetle Saturn, that's the defaults:

Beetle Saturn | YabaSanshiro
:-:|:-:
image | image
Yabause | Kronos
image | image

Same with Genesis Plus GX, PicoDrive, BlastEm, etc.

bslenul:
Oh yeah wtf, I was so sure they were all mapped like Genesis Plus GX/PicoDrive/etc. oO

My bad...

barbudreadmon:
Ok, let's go with that

barbudreadmon:
also added support for dinput

PR #811, opened 2022-01-02
android/dinput configs for 8BitDo Pro 2's new PID

landfillbaby:
based on previous configs
plus i got the triggers to be analog on android (not possible on windows due to driver differences)
hid driver just doesn't seem to work at all on any of my devices and i don't have a mac so i didn't bother
tested and ready for merge, but i might add some more drivers


landfillbaby:
ok i figured out why the previous configs weren't working. don't merge yet, fixing it

PR #813, opened 2022-01-04
Fixed syntax on PID &VID lines: Fixed Xbox Series controller cfg

Jaffacakelover:
* Some autoconfigs were missing quote marks around `vendor_id` and `product_id` values: Added them to match the majority.
* Added missing file extension to `xinput/XBOX Series Controller.cfg`.


RobLoach:
Thanks for the clean up!

atfrase:
Majority or no, are quotes actually correct for these fields? In the original docs linked from the README here (https://www.retroarch.com/index.php?page=controller-autoconfig), these two fields are shown without quotes since they're decimal values, not strings.

(Also, they're decimal and not hexadecimal, which I think may have been goofed in a few configs for devices that have hex IDs which coincidentally use only decimal digits such as the wired steam controller 0x1142; but that's a separate issue.)

Jaffacakelover:
I was going to say, if someone can detrmine which lines _don't_ want quote marks, I can go back through and remove from all files.

PR #816, opened 2022-01-09
Add HID support for Xbox Series X/S Controller

netmute:


netmute:
The new "Series" controller has a share button that I'm not really sure what to do with. I mapped it to exit RetroArch, because that seemed useful.

RobLoach:
Looks great!

> has a share button that I'm not really sure what to do with

In terms of that Share Button, is "Load State" an option? I imagine using that more than exit.

PR #827, opened 2022-02-21
Add 8BitDo Pro 2 udev config

claudiosanches:
I had some issues with my 8BitDo Pro 2 controller getting recognized as a 8BitDo M30 controller in RetroArch, so I decided to make a proper file to map this device. I hope it helps!

### DIFF
--- a/udev/8BitDo_Pro2_BT_X.cfg
+++ b/udev/8BitDo_Pro2_BT_X.cfg
@@ -1,0 +1,68 @@
+# 8BitDo Pro 2        - https://www.8bitdo.com/        - https://www.8bitdo.com/pro2/
+# Firmware v1.05      - https://support.8bitdo.com/    - https://support.8bitdo.com/firmware-updater.html
+# This is with the device started in Xinput mode.
+
+input_driver = "udev"
+input_device = "8BitDo Pro 2"
+input_device_display_name = "8BitDo Pro 2"
+
+# Hex vid:pid is found using "dmesg -w" or "tail -f /var/log/syslog" and converted to Decimal using http://www.binaryhexconverter.com/hex-to-decimal-converter
+# Hex vid:pid = 045E:02E0 -> Decimal vid:pid = 1118:736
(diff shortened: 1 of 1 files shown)


udev/8BitDo_Pro2_BT_X.cfg
```diff
@@ -0,0 +1,68 @@
...
+# This is with the device started in Xinput mode.
+
+input_driver = "udev"
+input_device = "8BitDo Pro 2"
+input_device_display_name = "8BitDo Pro 2"
+
+# Hex vid:pid is found using "dmesg -w" or "tail -f /var/log/syslog" and converted to Decimal using http://www.binaryhexconverter.com/hex-to-decimal-converter
+# Hex vid:pid = 045E:02E0 -> Decimal vid:pid = 1118:736
+input_vendor_id = "1118"
+input_product_id = "736"
```
RobLoach, line 12: Looks like there's a conflict across three other configs....
- https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/8BitDo_M30_BT_X.cfg
- https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/8Bitdo_Pro_SF30_BT_X.cfg
- https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/8Bitdo_Pro_SN30_BT_X.cfg

RobLoach (changes requested):
Looks like udev has three copies of these vendor pids. Not quite sure what the appropriate approach is here though.

udev/8BitDo_Pro2_BT_X.cfg
claudiosanches: I had the same thought, but I tested with a Pro 2, M30 and Pro SF30 4 controllers and works with the correct config in RetroArch.

claudiosanches: Thanks!

PR #829, opened 2022-02-28
Autoconfig for Atari VCS Modern Controller

ooshlablu:
This config allows the [Atari VCS Modern Controller](https://atarivcs.com/atari-vcs-wireless-modern-controller/) to be autoconfigured in retroarch. This configuration works for both bluetooth and USB connections.


RobLoach:
:fire:

PR #831, opened 2022-03-31
Add Retrode 2 USB adapter

revvv:
Tested on Wii and Wii U.

I provided dinput and udev files already some years ago.


bslenul:
Looks like you forgot to move it in the "hid" folder ;)

PR #835, opened 2022-04-27
Aya Neo Controller configuration

pastaq:
The Aya Neo line of handheld devices comes with extra buttons that act as keyboard macros. In order to utilize these buttons on Aya Neo devices under Linux I created the package [aya-neo-fixes](https://github.com/ShadowBlip/aya-neo-fixes) that creates a UInput device using python-evdev and overrides the built in xbox360 controller. Unfortunately this breaks compatibility with other xbox360 style configurations for this device. This configuration file supports the virtual Aya Neo Controller. I'm currently installing it with the aya-neo-fixes install script, but would prefer not to touch libretro files during install of my package. The package has been included in v33 of ChimeraOS, which has multiple Aya Neo users. Its also available in the AUR and on github. This patch will enable use of the built in controller for any Aya Neo users utilizing this patch, and will not conflict with the users who do not as the controller name is unique to the package.


offalynne:
@pastaq hi, would you please consider authoring a map for the SDL Community gamepad data repo as well ? we're missing data for this specific device https://github.com/gabomdq/SDL_GameControllerDB

thanks :-)

RobLoach (changes requested):
This is great! Would you mind adding the labels to this too?

RobLoach:
:rocket:

PR #840, opened 2022-06-07
Add Stratus Duo for Xinput, Nimbus+ and Tomee SNES/NES Adapters for HID/Dinput

stevenjamescox:
- Adds XInput config for the SteelSeries Stratus Duo via its included USB 2.4 GHz receiver.
- Adds HID controller config for the SteelSeries Nimbus+ in macOS.
  - The biggest benefit of this config is that both vertical axes will now be properly configured for users of this controller. In HID mode, MFi controllers' sticks' vertical axes are inexplicably inverted by default.
  - There are no D-Pad mappings in this config because RetroArch is unable to detect its D-Pad HID output in macOS. Other applications _are_ able to detect and use its D-Pad HID outputs. And RetroArch is able to detect _other_ MFi controllers' D-Pads.
- Adds DirectInput controller config for the SteelSeries Nimbus+ in Windows.
   - There are no L2 & R2 mappings in this config because its L2 and R2 buttons are not detected in RetroArch in Windows despite them working fully in the Windows Control Panel's "HID-compliant game controller" Properties interface.
- Adds small additions to the existing HID config for the SteelSeries Nimbus.
  - Friendly name, labels for L & R.
- Adds both DirectInput and HID configs for the Tomee NES and SNES Controller Adapters
  - These respective adapters connect original NES and SENS controllers to devices via USB.
- Adds HID config for the Hyperkin N64 Controller Adapter in PC Mode


offalynne:
Suggest commenting out `input_device` since that's the same name as every DInput device over Bluetooth on Windows

offalynne:
@stevenjamescox tangential to this PR, might you consider authoring maps for the Tomee NES & SNES adapters on MacOS for SDL ? see https://github.com/gabomdq/SDL_GameControllerDB

stevenjamescox:
@offalynne Oh for sure—will do!
1. I'm curious. Do you know how RetroArch handles the existence of matching profiles for different drivers for the same controller? Like if a there are `sdl2` and `hid` profiles for the same controller, when a user plugs it in, which profile does RetroArch use and why?
2. I only see `hid` as an option RetroArch for macOS -> Settings -> Drivers -> Controller. Are there plans for `sdl2` to be added to the macOS build soon? Maybe I can tweak build settings in Xcode to get `sdl2` into a local macOS build. Alternatively, I can probably create and test `sdl2` profiles over on my Windows machine.

offalynne:
I don't know how the `sdl2` driver works in RetroArch, nor which platforms or releases of RetroArch actually use the `sdl2` mappings. My interest re. SDL is unrelated to RA configuration since I personally have never found myself using the `sdl2` driver.

I rely on the community DB linked in my previous comment (which I co-maintain) for projects unrelated to RetroArch and just happen to keep an eye on this (the RA autoconfig repo) because the larger project of mapping devices is of mutual interest between the repos (besides enjoying using RA myself of course !)

To clarify, I'm asking if you're willing/able to contribute an SDL mapping ([tool here](http://www.generalarcade.com/gamepadtool/)) to the other repo however convenient: either a PR there, or just a comment wherever convenient for you with the mapping pasted so I can push that data upstream with credit.

stevenjamescox:
@offalynne Oh I see! Thank you for the links & information. I believe I used them correctly [here](https://github.com/gabomdq/SDL_GameControllerDB/pull/583).
I'll see what other controllers I have around that are missing from the database.

PR #844, opened 2022-06-23
Added 8BitDo Lite 2 and Lite SE

8BitDo:


RobLoach:
Thank you so much for pushing these forwards :+1: .... Really been enjoying these 8bitdo controllers.

PR #861, opened 2022-09-04
(Controller Autoconfig) Hori Fighting Edge PS4 Arcade Stick (xinput, dinput, sdl2)

oRuin:
Hori Fighting Edge PS4 Arcade Stick (xinput, dinput, sdl2)
[autoconfig.zip](https://github.com/libretro/retroarch-joypad-autoconfig/files/9485153/autoconfig.zip)


LibretroAdmin:
The PR as is unmergeable but we could perhaps just take your zip file and manually commit the contents of that.

PR #891, opened 2023-03-21
HandyGCCS: Fix the right analog stick (udev)

alkazar:
The right analog stick was inverted in both axes for the HandyGCCS handheld controller. This fixes it.

Tested on:
 - OneXPlayer Mini AMD
 - Aya Neo Air
 - Aya Neo 2021
 - Aya Neo 2


offalynne:
@alkazar Unrelated to this PR: can you check with SDL gamepadtool (https://www.generalarcade.com/gamepadtool/linux/) and see if
1. you get a unique GUID (seems like the intent of the virtual device is to spoof and Xbox pad)
2. if so, create a mapping per onscreen prompts and share ?

thanks

alkazar:
@offalynne Here is the mapping that was generated:

```
030003f05e0400008e02000010010000,X360 Controller,a:b0,b:b1,x:b2,y:b3,back:b6,guide:b8,start:b7,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,platform:Linux,crc:f003,
```

I am not sure if it is a unique ID.

offalynne:
Thank you !

PR #892, opened 2023-03-27
Android: 8BitDo Ultimate Wireless Controller

8BitDo:
Requires firmware update to v2.00 to enable bluetooth connectivity.


RobLoach:
Thank you! Love this controller.

PR #893, opened 2023-04-12
DragonRise NES, Fix conflict, Add USB NES RetroPort v2 support, fix conflict with Mayflash Arcade Stick Console Mode (udev), Fix conlict for Sony Dualshock 3

mechevarria:
Added an inactive configuration for a generic USB NES controller. The `input_vendor_id,` `input_product_id` and `input_device` currently match the values in `Retrolink_Sega_Saturn_USB_GamePad.cfg` but the buttons do not map correctly in Retroarch. This configuration is commented out by default to prevent conflicts

Link to the product
https://www.amazon.com/dp/B07M7SYX11?psc=1&ref=ppx_yo2ov_dt_b_product_details

~~Question:
Is there a more elegant way to switch between all the duplicate controller identities besides commenting/uncommenting the only one you use?~~

**UPDATE**
Also added support for [USB NES RetroPort v2](https://www.retrousb.com/product_info.php?cPath=21&products_id=158)
Also **fix** conflict with [MayFlash F300/F500 Arcade stick](https://www.mayflash.com/product/mayflash_arcade_stick_f500_v2.html) console mode
Also **fix** conflict with Sony Dualshock 3.

Several files have the same vendor/product. Use one to prevent conflicts and comment to the location of the one in use so you can quickly comment/uncomment for the controller that you have


hizzlekizzle:
Not currently, no. Having some way to pick among a list of matches would be ideal but isn't implemented yet

PR #910, opened 2023-06-08
Update Nintendo Switch Pro Controller (old).cfg

davidhedlund:


davidhedlund:
Also rename `sdl2/Update Nintendo Switch Pro Controller (old).cfg` to `sdl2/Update Nintendo Switch Pro Controller (bare).cfg` to make it consistent with `udev/Update Nintendo - Switch Pro Controller (bare).cfg`

PR #913, opened 2023-06-25
Added Mayflash MF103 (N64 to USB, Swtich/Windows) config

Matsilagi:
Modern versions of the adapter didn't seem to have a preset, so i made one. Tested with a off-brand N64 controller, with a up-to-date MF103 (V1.00.22.08 as the time of writing this) firmware. But should work with any N64 controller that the adapter supports.

Make sure to follow the mode instructions inside each config, but incase you can't read it for any reason: Recommendations are keeping it on DInput (Blue LED mode) for most of the use-cases, only use the Green LED mode (XInput) as a last-resort / fallback.

Platforms tested:
PC - Windows 11 x64, RetroArch 1.15.0 (both x86 and x64)
Android - Samsung Galaxy S20 FE (Running RetroArch AArch64, 1.15.0)

Cores tested:
Mupen64PlusNext (Works fine)
ParaLLEl (Recommended and most tested core)
ParaLLEl-Next (Also recommended and most tested core, copied the core file manually to RetroArch standalone from Parallel Launcher)


xinput/Controller (MAYFLASH N64 USB Adapter MF103).cfg
```diff
@@ -0,0 +1,56 @@
+# Mayflash N64 Controller Adapter (MF103) for Nintendo Switch and Windows
+# Adapter mode on either DInput (BLUE Led) or XInput (GREEN Led) mode works.
+# Calibrate in Windows / Linux for full analog range
+# Analog to Digital Type = Right Analog
+# configured for ParaLLEl (Remap might be needed for Mupen64Plus)
+
+input_driver = "xinput"
+input_device = "Controller (MAYFLASH N64 USB Adapter MF103)"
+input_vendor_id = "12068"
+input_product_id = "244"
```
RobLoach, line 10: This ends up being a duplicate to the other xinput one. Is having duplicate entries here intended?

RobLoach (changes requested):
Thanks so much! Just a quick question about Controller.cfg

xinput/Controller (MAYFLASH N64 USB Adapter MF103).cfg
Matsilagi: I think so, its a single adapter with 2 ports and multiple mdoes built-in, so, if you have 2 of them running with the same mode, it won't give a problem.

PR #914, opened 2023-07-12
Add Sony DS4 USB Wireless Adaptor

superjamie:
This is a USB device which does the Bluetooth pairing in hardware so appears as a regular USB controller.

https://blog.playstation.com/archive/2016/08/22/introducing-the-dualshock-4-usb-wireless-adaptor/

It has profiles in dinput and hid, but not udev. Add udev.


RobLoach:
👍

PR #915, opened 2023-07-26
Update and rename Nintendo - Switch Pro Controller (bare).cfg to Nint…

davidhedlund:
…endo - Switch Pro Controller (old).cfg

To make it consistent with https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/sdl2/Nintendo%20Switch%20Pro%20Controller%20(old).cfg


davidhedlund:
Thank you!

PR #922, opened 2023-09-02
Add support for PowerA NSW Wired Controller

mwprado:
Add support for PowerA NSW Wired Controller


hizzlekizzle:
I suppose there's no need for button labels, since they already match the retropad?

Good call mapping the light gun trigger. We should probably do that on every profile.

Would it make sense to map the 'home' button to menu_toggle?

mwprado:
>

mwprado:
> I suppose there's no need for button labels, since they already match the retropad?
> 
> Good call mapping the light gun trigger. We should probably do that on every profile.
> 
> Would it make sense to map the 'home' button to menu_toggle?

Hi, it's my first pull request. I'm trying learn github's interface.
1 - If I understood, there's no need for button labels. Below is m controller model: 
https://www.powera.com/p/nintendo/nintendo-switch/controllers/wired/enhanced-wired-controller-for-nintendo-switch-woo-hoo-mario-nsgp0001-01/

2 - I don't know what standard trigger button used to be. What button do you suggest? I will submit another file. 

3 - I mapped the controller's home button. It's working, but there isn't  any difference in file. Do I have to submit any other configuration file? 

PowerA-NSW-Wired-Controller

hizzlekizzle:
2. it's already in the one you submitted. We haven't included it in profiles historically, but I think it probably makes good sense to do so in the future.
3. Yeah, you have to add it in manually with a text editor. It won't save in there, AFAIK.

mwprado:
> 2. it's already in the one you submitted. We haven't included it in profiles historically, but I think it probably makes good sense to do so in the future.
> 
>     3. Yeah, you have to add it in manually with a text editor. It won't save in there, AFAIK.

I've added home button.

PR #926, opened 2023-10-01
(SDL2) Add ClockworkPi DevTerm

xordspar0:
Please be patient with me as I'm just figuring this out.

1. I only provided config for sdl2 because I have only figured out how to configure this device with sdl2 so far.
2. Is there a way to enable analog to digital conversion by default? This controller has a d-pad that's wired up as a pair of analog axes.


hizzlekizzle:
you should be able to just assign them to retropad-up/down/left/right instead of the retropad analogs, regardless of whether the pad sees them as analog or digital.

xordspar0:
I tried that by doing this:

```diff
- input_l_y_minus_axis = "-1"
- input_l_y_plus_axis = "+1"
- input_l_x_minus_axis = "-0"
- input_l_x_plus_axis = "+0"
+ input_up_btn = "-1"
+ input_down_btn = "+1"
+ input_left_btn = "-0"
+ input_right_btn = "+0"
```

But that didn't work. That made my d-pad do nothing. Retroarch doesn't seem to recognize axes as an option for mapping to buttons.

hizzlekizzle:
It would be input_*_axis rather than input_*_btn in this case.

xordspar0:
Ah, that works perfectly, thanks!

PR #930, opened 2023-10-06
Add dinput config for 8BitDo N64 Modkit

bcat:
The [8BitDo N64 Modkit](https://shop.8bitdo.com/products/8bitdo-mod-kit-for-original-n64-controller) is a replacement PCB and "rumble pak" that converts an OEM N64 Controller to a USB/Bluetooth controller with two modes: one that emulates an NSO N64 controller (for which there are already autoconfig files), and the other that presents as a normal DirectInput controller with a unique PID/VID. This pull request adds a `dinput` config file for the second mode of operation.

I tried to the match the formatting of the other 8BitDo `dinput` configs, but since the controller physically offers the same buttons as the NSO N64 Controller (including addition of ZR and Home buttons), I used the same RetroPad mappings as the existing `N64_Nintendo_Switch_Online.cfg` (with DirectInput button/axis numbers changed appropriately).

I tested this on the Mupen64Plus-Next core, and it seems to work fine.

Note that I don't love the compromises these mappings (and the equivalent NSO N64 mappings) make for the controller to work with both default Mupen config and RetroArch menus. In particular, double-binding the controller's physical "B" button to enable menu navigation is a hack, and the end result still leaves menu OK/cancel flipped vs. a normal controller. But honestly, the RetroPad abstraction doesn't work very well with controllers whose face buttons aren't laid out like a SNES or DualShock pad, and I suspect the NSO controller mappings are the least bad choice to make under the circumstances.

I don't have a Linux setup running RetroArch right now, so I didn't add a corresponding `udev` config, though it should be easy for someone else to do so if they want.

### DIFF
--- a/dinput/8Bitdo_N64_Modkit.cfg
+++ b/dinput/8Bitdo_N64_Modkit.cfg
@@ -1,0 +1,76 @@
+# 8BitDo N64 Modkit           - http://www.8bitdo.com/     - https://shop.8bitdo.com/products/8bitdo-mod-kit-for-original-n64-controller
+# Firmware v1.00              - http://support.8bitdo.com/
+# This is with the device started in Android (D-Input) mode.
+# Note: USB and Bluetooth connections have the same PID.
+# Note: Switch mode uses the NSO N64 controller's VID/PID instead.
+
+input_driver = "dinput"
+input_device = "8BitDo N64 Modkit"
+input_device_display_name = "8BitDo N64 Modkit"
+
+# Hex vid:pid and Decimal vid:pid is shown in the "log_verbosity" window, enable "log_verbosity" in retrorch.cfg and run RetroArch.
+# Hex vid:pid = 2DC8:2869 -> Decimal vid:pid = 11720:10345
+
+input_vendor_id = "11720"
+input_product_id = "10345"
+
+# The controller's physical layout matches the NSO N64 controller, and so the
+# button mappings and labels mirror N64_Nintendo_Switch_Online.cfg. However, the
+# D-Input button numbers differ (unless the controller is put in Switch mode).
+#
+# These mappings share some features and quirks with the NSO controller:
+#
+# 1. The mappings work out of the box with default Mupen64Plus-Next core config.
+# 2. The button labeled "B" is mapped to Y on the RetroPad (for Mupen)... and
+#    also to A on the retropad. This duplicate mapping is ignored in Mupen by
+#    default, but enables RetroArch menu navigation.
+# 3. The button labeled "ZR" is mapped to Select on the RetroPad, not R2. This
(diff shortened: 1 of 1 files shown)


RobLoach:
Looks great, while I don't have the modkit to build the config for udev, this should be good to go as a first step. Thanks!

SosPosos:
Hi, a and b was swapped for some reason.. and did you manage to make rumble work in switch mode?

bcat:
> Hi, a and b was swapped for some reason.

I agree this is a bit strange, but it's actually intentional (and matches the mapping for the official NSO N64 controller). If you look at the [default Mupen64Plus-Next controller conifg](https://docs.libretro.com/library/mupen64plus/#joypad-and-analog-device-type-table), you will see the following:

* RetroPad B -> Emulated N64 A
* RetroPad Y -> Emulated N64 B

This way, the physical positions of the N64 B and A buttons relative to one another are preserved on a SNES-style RetroPad.

But the abstraction gets confusing when the physical controller _doesn't_ have an SNES-style button cluster. In order for the N64 controller to work correctly with default Mupen core mappings, this autoconfig maps as follows:

* Physical N64 A -> RetroPad B -> Emulated N64 A
* Physical N64 B -> RetroPad Y -> Emulated N64 B

However, those two mappings would leave no way to press RetroPad A, which is kind of important for navigating the RetroArch menus. :) To work around this, we define an additional mapping (and, again, the official NSO controller's autoconfig file does the same thing):

* Physical N64 A -> RetroPad B -> Emulated N64 A
* Physical N64 B -> RetroPad Y -> Emulated N64 B
* Physical N64 B -> RetroPad A

With these mappings, the physical buttons perform the right actions in emulated games by default, and it's possible to navigate the RetroArch menus as well. But A and B are swapped when navigating menus (Physical N64 A -> RetroPad B, Physical N64 B -> RetroPad A).

IMO, this is a drawback of the RetroPad abstraction itself. I really wish libretro had support for controller "shapes" other than SNES/PSX style. In particular, native awareness of physical controllers with six physical face buttons would make button mapping for N64- and Genesis-style controllers _much_ nicer.

But that's not how RetroArch works today, and so the author of NSO controller config chose to prioritize default functionality at the expense of reversed buttons in the menus. That seemed like a reasonable choice, and I decided to follow it for the config that got checked in. :)

(Note that I _personally_ do not use this config as is. I only play N64 games on an N64-style controller, so I updated my autoconfig locally to map Physical N64 A -> RetroPad A and Physical N64 B -> RetroPad B, and then I adjusted my Mupen core control mappings accordingly. But that didn't seem like a good default config since the controller wouldn't be compatible out of the box with default Mupen mappings.)

> And did you manage to make rumble work in switch mode?

I did not test this, sorry. (I use my controller exclusively in Switch mode on Switch and DInput mode on PC.)

PR #936, opened 2023-10-13
Fix menu button bindings for 8BitDo SN30 Pro

bcat:
The menu button binding for this controller (in both the Bluetooth and USB configs) is wrong. I'm not sure if it's always been wrong, or if the button mapping has changed in recent firmware, but in any case, none of the buttons on the controller map to button 2. In fact, the menu button (8BitDo "heart" logo) is button 12.

Should I also update the udev configs? They seem to use the exact same mappings as D-Input, so it might make sense, but I don't have a way of testing with udev right now.


RobLoach:
> Should I also update the udev configs?

Quite possibly, yeah. Thanks!

bcat:
Updated the udev config too, but it's untested, so if you're able to double check it's correct, that's much appreciated!

PR #937, opened 2023-10-23
Add support to Retroflag SuperPack

impeeza:
Retroflag Xbox Series Controllers SuperPack is an add-on to a Xbox Series Controller, which allows to connect the controller via regular Bluetooth to a PC on X-input mode or to a Nintendo Switch Console.

Expanding the capabilities of the controller to regain Regular Bluetooth.

Also ads a rechargeable battery.


xinput/Retroflag Xbox Series Controller SuperPack.cfg
```diff
@@ -0,0 +1,62 @@
+# Adapted RetroFlag Xbos Series Controllers SuperPac
+# in "X" mode v�a Bluetooth
```
RobLoach, line 2:
``` suggestion
# in "X" mode with Bluetooth
```

impeeza: yeah, sorry, I am Hispanic, so my typo was write that word on Spanish. The Retroflag SuperPack have a switch which allows you to select Nintendo Switch or X-Input mode using traditional Bluetooth so you can use the XBox Series Controller on system without BLE.

Thanks for your help.

RobLoach:
Thanks so much!

PR #938, opened 2023-10-30
Compare Lakka Differences

RobLoach:
The following are the configs that Lakka overrides from the default joypad configs. Is there anything here that should be kept?


hizzlekizzle:
I don't see any harm in adding in the ones that don't conflict, but it looks like the conflicts are just rolling back to an older version. I'm guessing this is related to changes in the kernel driver that changed the mapping...?

bagelbyheart:
@RobLoach don't forget the 8bitdo Pro+ driver issue.

RobLoach:
Thanks for the ping, bagel. I think keeping the patches in Lakka is fine for now :shrug:

PR #939, opened 2023-10-30
Update 8BitDo_SN30_Pro+_BT_X.cfg

bagelbyheart:
Corrected udev mappings for the 8BitDo SN30 Pro+ in xinput bluetooth mode.

Previously input_b_button was mapped to the right controller button, etc, causing PlayStation games to have Square/Triangle and Cross/Circle swapped.

Settings for the BT_B.cfg were already correct.
Settings for the USB.cfg were already correct.


RobLoach:
I think I ran into this issue, and thought I was crazy. Thanks for assuring I'm not insane.

RobLoach:
@bagelbyheart As per https://github.com/libretro/retroarch-joypad-autoconfig/pull/936 , should the Menu Toggle button be 12 too?

bagelbyheart:
> @bagelbyheart As per #936 , should the Menu Toggle button be 12 too?

I'm not actually sure. I tried to check, and while Lakka is definitely recognizing the "Home" button on the controller, it won't let me map it to any input, instead it's treating it like backing out of input selection.

I was able to map it to "Hotkey Enable" where it appears as `(Key: mode)` and I don't see `mode` as the label to any of the inputs, so I'm a bit stumped.

bcat:
FWIW, I don't have one myself to check, but the SN30 Pro+ and the SN30 Pro are fairly different (e.g., the Pro+ has analog triggers and the Pro doesn't). The Pro+ is actually more like the Pro 2 in terms of features and appearance. (Not sure if that translates into similar button numbering or not.)

bcat:
Though I guess the Pro 2 also uses 12 for its dinput and udev menu buttons, so _maybe_ 12 is actually right for the Pro+ too?

Sanaki:
This commit screwed up the mappings for me. Are you sure you had your controller properly updated? They no longer match retropad positions with my controller.

RobLoach:
Which Mode are you using?

Sanaki:
Xinput, via X+start

bagelbyheart:
I'll check my controller firmware when I get home, looks like there's a firmware updater and the ultimate software might tell me my version. @Sanaki could you check the firmware version on yours as well?

Sanaki:
Firmware version on mine is 6.01. I know with other 8bitdo controllers mappings have changed on a few updates, the same may have happened with this one.

bagelbyheart:
I'll verify that is what's happening, if so it should be an easy revert, and I'll add a comment to the config file explaining for other people that might "fix" it down the line =)

bagelbyheart:
Started on firmawre v3.02. My lakka install still has the old mapping.

```cfg
input_b_btn = "0"
input_y_btn = "2"
input_a_btn = "1"
input_x_btn = "3"
```

Confirmed still acting up. If I test the control mapping within Valkyrie Profile Cross and Circle are swapped as are Triangle and Square.

I then updated to firmware v6.01 and tested again. I still see the swap between Cross/Circle and Triangle/Square.

@Sanaki, is it possible that you have a profile applied via the 8BitDo Ultimate software? Mine is set to default and basically demonstrates the same thing I'm seeing, in the software B=A and vice versa. I've attached a screenshot of my setup post upgrade.

image

PR #1070, opened 2023-12-12
Rename DualSense Wireless Controller (PS5).cfg to Sony DualSense Wire…

davidhedlund:
…less Controller (PS5).cfg

Sony


RobLoach:
Why are we adding `Sony` to the filename? Could you provide some context to some of the Pull Requets? Thanks.

davidhedlund:
> Why are we adding `Sony` to the filename? Could you provide some context to some of the Pull Requets? Thanks.

For the sake of consistency: "Sony " is added in the beginning for the other "DualSense" cfg files.

Also, it's difficult to overview find the files when they are not grouped this way.

PR #1071, opened 2023-12-12
Rename DualSense5.cfg to DualSense.cfg

davidhedlund:
DualSense not DualSense 5


davidhedlund:
Assumption. I don't have access to Windows.

PR #1091, opened 2023-12-28
Add dualsense edge controller and handheld daemon controller in udev

antheas:
Adds the the Dualsense Edge Controller config, the linux version of which is missing.

Adds the Handheld Daemon Controller config, which is a controller created by the Handheld Daemon to perform remapping for handheld devices (e.g., the Legion Go)
[github.com/antheas/hhd](https://github.com/antheas/hhd)

The Handheld Daemon also allows remapping to a DS5 Edge controller. And it uses a different name which is subject to change. For this reason, I left the real DS Edge name commented out in the DS Edge config.

If we find compatibility issues with this approach we might switch to the original controller name.


udev/Handheld Daemon Controller.cfg
```diff
@@ -0,0 +1,30 @@
...
+input_l_x_plus_axis = "+0"
+input_l_x_minus_axis = "-0"
+input_l_y_plus_axis = "+1"
+input_l_y_minus_axis = "-1"
+input_r_x_plus_axis = "+3"
+input_r_x_minus_axis = "-3"
+input_r_y_plus_axis = "+4"
+input_r_y_minus_axis = "-4"
+input_gun_trigger_mbtn = "1"
+input_gun_offscreen_shot_mbtn = "2"
```
RobLoach, line 30: This is great, thanks... Would you mind adding the labels here too?

RobLoach (changes requested):

antheas:
I will review the playstation file and the handheld daemon file in a fiew days. There might be some changes with the vid for handheld daemon.

I copied the names for the DSE from its dinput version. I will reference the normal dualsense to add the rest of the names. Should I add them to the dinput version too?

RobLoach:
Good idea. Thanks so much for the additions here

antheas:
It seems that gun trigger and gun button are not part of the configuration, but were added by retroarch by mistake. I notice that it adds them by mistake.

Can you raise a separate issue for this where appropriate?

antheas:
I removed the extra buttons, and fixed the menu button by copying it from the other dualsense configuration.

Also, it turns out that the new linux driver adds a 'DualSense" to the normal DualSense name as well. So the normal dualsense will not work in linux with that name. So I added a copy of the configuration for that as well.

I checked name, vid, and pid match. And that retroarch at least loads the emulated (non-edge) version.

You may rebase and merge if there are no other issues.

PR #1092, opened 2023-12-29
sdl2: Add 8BitDo Pro 2

xordspar0:
There's already at least one mapping for this controller for udev, but udev doesn't currently work in the Flatpak version of RetroArch.

Maybe this is well known, but just to clarify, I tried this device in both USB and Bluetooth mode and it showed up the same to SDL. The vendor and product IDs were the same, and RetroArch detected the controller in both modes.


RobLoach:
Thanks! Have been really curious about porting the controller db over to our format here: https://github.com/gabomdq/SDL_GameControllerDB

PR #1109, opened 2024-01-27
Add Logitech Driving Force GT

ltrvs:
Playstation face buttons assigned as usual.

Wheel assigned to left analog's left+right.
Throttle pedal assigned to right analog's up.
Brake pedal assigned to right analog's down.

Analogs mapped for PS1 games like _Gran Turismo 2_ and  _Toca 2_. Are analog settings OK like this or should they be removed to let users map as they see best?

https://en.m.wikipedia.org/wiki/Logitech_Driving_Force_GT

Update: renamed the file to have "Logitech" prefix. Did the same for G27 config also and added a display name for it.

(The G27 config probably needs more love, see https://forums.libretro.com/t/g27-wheel-causes-ui-control-issues/20042 . I don't have the G27 device, so won't change the button config based only on forum comments.)


ltrvs:
Set to draft, will add missing _input_menu_toggle_btn_

PR #1112, opened 2024-02-16
8BitDo SN30 Pro+ Bluetooth X button mapping

RobLoach:
@bagelbyheart @Sanaki Something up with this config? Both are on latest firmware and on Xinput mode? I'm unable to test as I don't have a SN30 Pro+, nor bluetooth

Reverts libretro/retroarch-joypad-autoconfig#939


bagelbyheart:
I'll be checking this shortly to verify. If this is the case, lets add a comment to the top of the file calling out the firmware version.

bagelbyheart:
Started on firmawre v3.02. My lakka install still has the old mapping.

```cfg
input_b_btn = "0"
input_y_btn = "2"
input_a_btn = "1"
input_x_btn = "3"
```

Confirmed still acting up. If I test the control mapping within Valkyrie Profile
Cross and Circle are swapped as are Triangle and Square.

I then updated to firmware v6.01 and tested again. I still see the swap between
Cross/Circle and Triangle/Square.

@Sanaki, is it possible that you have a profile applied via the 8BitDo Ultimate
software? Mine is set to default and basically demonstrates the same thing I'm
seeing, in the software B=A and vice versa. I've attached a screenshot of my
setup post upgrade.

image

Sanaki:
I have never opened the ultimate software because it's incompatible with Linux, so there's no way I could have applied such a profile. Check against the gamepad tester in a chromium based browser, see what behavior you see there.
https://hardwaretester.com/gamepad

RobLoach:
Thanks for the investigation here, both of you!

PR #1115, opened 2024-02-29
sdl: Fix PS5 Controller Product ID

davidhedlund:
replaced
input_product_id = "1476"

with
input_product_id = "3302"

Reference:
* RetroArch Flatpak uses SDL2 by default and generated "input_product_id = "3302"" via "Settings -> Input -> Port 1 Binds -> Save Controller Profile"
* `input_product_id = "1476"` is the product ID for [PS4 Controller.cfg](https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/sdl2/PS4%20Controller.cfg).


davidhedlund:
`input_product_id = "1476"` is for DualShock 4

RobLoach:
Thanks

PR #1119, opened 2024-03-08
add sdl support for NSO controllers

geogolem:
note .. in order to take advantage of these, you need to replace the SDL2.dll that comes with retroarch with a newer version? Why is retroarch on such an old version anyway?

Using the latest SDL2.dll along with these autoconfiguration files is a good solution to https://github.com/libretro/RetroArch/issues/13520


geogolem:
I noticed that the vendor ID etc. are strange in these changes I made.. Nintendo should be 1406.. and it is for me in linux.. but in Windows it is showing up as 3... I am not sure what the issue is.. if anybody has any idea please let me know..

RobLoach:
Good call out on older SDL. I'm unsure why it packages an older one. Worth bringing this update in. Thanks!

PR #1127, opened 2024-04-25
udev: Add RetroUSB N64 RetroPort

mechevarria:
Add mapping for [USB N64 Retroport](https://www.retrousb.com/product_info.php?cPath=21&products_id=82&osCsid=d1511c0e801bd16feddd987fb03be95f). Verified there are no conflicting vendor and product pids


offalynne:
what's up with the x and y button mappings ?

PR #1128, opened 2024-05-05
udev: Add mapping for Xbox controllers using the xpadneo driver

CosmicHeron:
Added a profile for Xbox Controllers using the [xpadneo](https://atar-axis.github.io/xpadneo/) driver on Linux


RobLoach:
Thanks!

PR #1137, opened 2024-06-07
[DInput] Add Retro Fighter Brawler64 support

Cosmo480:
Retro Fighters Brawler64 Wireless Gamepad
UPC: 850046080067

The physical B button shares 2 functions: in game, it serves as the n64 "B" button. in the retroarch menus it serves as a back button, ("A" according to port 1 control menu).  This is the best way i could figure out how to map this gamepad for nintendo 64 games. Only problem is that the "Y" button is also serves for deleting hotkeys in the menu. work around for this is to temporarily set "y" to a C up, down, left, or right button. remove the hotkey errors and then resetting to default controls afterwards in port 1 control.

### DIFF
--- a/dinput/Wireless Gamepad.cfg
+++ b/dinput/Wireless Gamepad.cfg
@@ -33,7 +33,7 @@ input_r_y_minus_btn = "11"
 
 input_menu_toggle_btn = "12"
 
----------------------------------------
+# Labels
 
 input_b_btn_label = "A"
 input_y_btn_label = "B"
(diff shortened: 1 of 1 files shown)


dinput/Wireless Gamepad.cfg
```diff
@@ -0,0 +1,65 @@
...
+input_l_y_minus_axis = "-1"
+
+input_r_x_minus_btn = "3"
+input_r_x_plus_btn = "10"
+input_r_y_plus_btn = "2"
+input_r_y_minus_btn = "11"
+
+input_menu_toggle_btn = "12"
+
+---------------------------------------
```
RobLoach, line 36:
``` suggestion
# Labels
```

PR #1144, opened 2024-07-14
android: Fix DualSense Wireless Controller name

davidhedlund:
* Correct generated file name


offalynne:
The `input_device` field is incorrect and no longer matches after this change. This is a regression and should be reverted.

davidhedlund:
> The `input_device` field is incorrect and no longer matches after this change. This is a regression and should be reverted.

How can you be certain about that?

In my experience, the name "DualSense Wireless Controller" appears when the controller is connected via Bluetooth to Android versions 10 through 13. Conversely, the name "Wireless Controller," as you suggest, is associated with the Sony DualShock 4 v2 when connected over Bluetooth. However, the input_device is not relevant for the Android controller driver. I reported this issue two hours ago:https://github.com/libretro/RetroArch/issues/16920.

PR #1146, opened 2024-07-18
dinput: Fix Brawler64 config matching Switch Pro Controllers

### LINKED ISSUE #1145: Nintendo Switch Pro Controller Regression
The latest version of autoconfig causes Nintendo Switch Pro controllers to be misidentified as "Retro Fighter Brawler64 Wireless Gamepad".

This is almost definitely caused by #1137, but I'm not 100% sure on why. Likely though its just because they share the exact same vendor and product ID, but I can't verify this as I don't own any Brawler64 controllers. IMO RetroArch shouldn't ship with a config which breaks common controllers so widely.

Lustyn:
Fixes #1145


RobLoach:
Seems like an appropriate fix. Switch Pro controllers will be more popular.

PR #1150, opened 2024-07-20
[READY] android: Update `DualSense Wireless Controller.cfg`

davidhedlund:


RobLoach:
The only change here is whitespace. Thanks, but naw.

PR #1156, opened 2024-07-29
[READY] sdl2: Update `Nintendo Switch Pro Controller.cfg`

davidhedlund:
* Updated the bindings
  - Fixes https://github.com/libretro/retroarch-joypad-autoconfig/issues/1204
* Streamlined the autoconfig file by removing comments and blank lines, sorting variables
* Added `input_screenshot_btn`/`input_screenshot_label` for the unused screenshot button.
* The screenshot button does generate any binding when pressed, so I didn't set `input_screenshot_btn`/`input_screenshot_label` for the unused screenshot button like I did for udev/linuxraw.


davidhedlund:
Please, let's quickly solve this issue first:
* https://github.com/libretro/retroarch-joypad-autoconfig/issues/1157

sdl2/Nintendo Switch Pro Controller.cfg
```diff
@@ -1,35 +1,23 @@
...
-# 
-# The easiest solution for most users is to install the nintendo-hid DKMS module, available from https://github.com/nicman23/dkms-hid-nintendo/
-
 input_driver = "sdl2"
 input_device = "Nintendo Switch Pro Controller"
-
 input_vendor_id = "1406"
 input_product_id = "8201"
-
-input_a_btn = "1"
```
RobLoach, line 16: Looks like we're flipping A/B. RetroArch uses A for right face button, and B for the bottom face button. Is this change correct? You can see a visual of it over at:

RobLoach (changes requested):

sdl2/Nintendo Switch Pro Controller.cfg
davidhedlund: This is correct (just generated a test autoconfig to verify it for you), it's generated by RetroArch 1.19.1.

I am confident that every single binding in all pull requests is 100% correct. The only manual additions I've made are for the analog L2/R2 buttons on PlayStation controllers, specifically for drivers that didn't generate them in RetroArch. I maintain a detailed spreadsheet that tracks all controllers, drivers, manual bindings, and bug reports.

All autoconfigurations I've uploaded since the release of RetroArch 1.19.1 have been generated from that version. I've tested every autoconfig 2-3 times.

I'm certain that all pull requests can be approved without the need to merge them and potentially review everything again. I would prefer not to spend additional time on this, as it has already taken a significant amount of effort to reach this point.

davidhedlund: > Looks like we're flipping A/B. RetroArch uses A for right face button, and B for the bottom face button.

Since which RetroArch version (please give me a reference link if possible)?

* My Android autoconfigs have been generated with RetroArch 1.17.0 due a bug that prevents me from installing the current version (1.19.1). I submitted that bug here: https://github.com/libretro/RetroArch/issues/17014

davidhedlund:
> @davidhedlund The mappings are good, but the labels for L2 and R2 are messed up. L2 and R2 are considered axes, but the labels are for L2 and R2 buttons, so they do not show up in RetroArch.

I just submitted a PR:
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1248

PR #1159, opened 2024-08-01
[READY] linuxraw: Create `Nintendo Switch Pro Controller (non-HID) (default-off).cfg`

davidhedlund:


RobLoach (changes requested):
Could you clarify why we are adding this one when `linuxraw/Nintendo Switch Pro Controller.cfg` already exists?

It looks like the only difference is the A/B flipped. RetroArch uses the retropad scheme:

davidhedlund:
> Could you clarify why we are adding this one when `linuxraw/Nintendo Switch Pro Controller.cfg` already exists?
> 
> It looks like the only difference is the A/B flipped. RetroArch uses the retropad scheme: https://www.lakka.tv/doc/Input-settings/

I added "(non-HID) " to the filename

davidhedlund:
> Could you clarify why we are adding this one when `linuxraw/Nintendo Switch Pro Controller.cfg` already exists?

I added "(non-HID) " to the filename for this PR.

> It looks like the only difference is the A/B flipped. RetroArch uses the retropad scheme: https://www.lakka.tv/doc/Input-settings/

That is incorrect. The HID file have completely different bindings. Have a look at: https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/linuxraw/Nintendo%20Switch%20Pro%20Controller.cfg

PR #1160, opened 2024-08-01
[READY] sdl2: Update and rename `Nintendo Switch Pro Controller (old).cfg` to `Nintendo Switch Pro Controller (non-HID) (default-off).cfg`

davidhedlund:
* Added `input_device_display_name` to make the system compatibility clear.
* Removed comments


sdl2/Nintendo Switch Pro Controller (default-off).cfg
```diff
@@ -1,19 +1,8 @@
...
-# 
-# The easiest solution for most users is to install the nintendo-hid DKMS module, available from https://github.com/nicman23/dkms-hid-nintendo/
-
 input_driver = "sdl2"
-input_device = "Nintendo Switch Pro Controller (old)"
-
-# input_vendor_id = "1406"
-# input_product_id = "8201"
-
+#input_device = "Nintendo Switch Pro Controller"
```
RobLoach, line 2: The only thing we're really changing here is removing the `input_device`. Could you clarify why we're making this change? Thanks a lot.

RobLoach (changes requested):

PR #1161, opened 2024-08-01
[READY] udev: Update `Nintendo Co., Ltd. Pro Controller (default-off).cfg` and create `Nintendo Switch Pro Controller (default-off).cfg` / `Pro Controller (default-off).cfg` copies

davidhedlund:
Organize variables in accordance with the most recent RetroArch version. This reorganization will facilitate easier comparison between autoconfig files.


davidhedlund:
I will create a group of pull requests that include both HID and non-HID auto-configurations for linuxraw.

PR #1158, opened 2024-08-01
android/Pro Controller.cfg: input_menu_toggle_btn comment

davidhedlund:
input_menu_toggle_btn is bound to the Capture button (Circle label) instead of the the Home button (Home label), because the Home button is bound to Android's Home button.

PR #1162, opened 2024-08-01
android: Update Pro Controller.cfg

davidhedlund:
Organize variables in accordance with the most recent RetroArch version. This reorganization will facilitate easier comparison between autoconfig files.


davidhedlund:
Already sorted in https://github.com/libretro/retroarch-joypad-autoconfig/pull/1158/files

PR #1164, opened 2024-08-02
[READY] linuxraw: Create `Nintendo Switch Pro Controller.cfg`

davidhedlund:


davidhedlund:
I will create a group of pull requests that include both HID and non-HID auto-configurations for linuxraw.

PR #1165, opened 2024-08-05
dinput: Created autoconfig mapping for Hyperkin Scout Premium Wireless BT and Retro-Bit Saturn Bluetooth Controller

jedinjapan:
Additional mapping file for the Hyperkin Scout Premium Wireless BT. Prevents conflicts with other "Wireless Gamepad" devices.


hizzlekizzle:
Thanks for the contribution!

Could we trouble you to add button labels, since you have the physical controller available? This will allow RetroArch to show the labels the controller uses in menus instead of the generic retropad aliases. You can see an example of it here: https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/Xbox_One_Controller.cfg#L33

jedinjapan:
Sure, I'll go through and do this. 

I have an additional PR for a Retro-bit Bluetooth Saturn Gamepad as well that I will do the same.

jedinjapan:
Labels added to both added configurations. Tested in bsnes and Genesis Plus GX and labels appear as intended.

PR #1176, opened 2024-08-10
[READY] linuxraw: Create DualSense Wireless Controller.cfg

davidhedlund:


davidhedlund:
"DualSense Wireless Controller" does not exist for any Linux kernel versions. I think I simply confused the controller driver with android that uses "DualSense Wireless Controller".

PR #1182, opened 2024-08-20
[READY] Update and rename `PS4 Controller.cfg` to `PS4 Controller (v1).cfg`

davidhedlund:
* removed `input_device`. See https://github.com/libretro/RetroArch/issues/16920
* Renamed to `PS4 Controller (v1).cfg` to make room for `PS4 Controller (v2).cfg`, which can be renamed to `PS4 Controller.cfg later`, as it is a newer model commonly sold in stores.


sdl2/PS4 Controller (v1).cfg
```diff
@@ -1,5 +1,4 @@
 input_driver = "sdl2"
-input_device = "PS4 Controller"
```
RobLoach, line 2: Most configs have this `input_device` field defined. Any reason to not have the descriptor in there?

PR #1192, opened 2024-09-15
android: Update Xbox 360 Controller configs

landfillbaby:


landfillbaby:
i just noticed something. in the config for the wired version:

```
input_b_btn_label = "A"
input_y_btn_label = "X"
[...]
input_a_btn_label = "B"
input_x_btn_label = "Y"
```

but in the wireless adapter version, they're not swapped like that.

this is obviously due to nintendo/xbox differences. but which one is right?

hizzlekizzle:
I would assume the swapped version is correct from a physical layout standpoint, which is typically how we handle things.

landfillbaby:
ready for merge :)

landfillbaby:
oh just realised this could be merged using https://github.com/libretro/RetroArch/pull/16990

landfillbaby:
```md
this is the proper footnote[^1] syntax.

[^1]: or reference.
```

RobLoach:
Since we're making updates here, I think it would be appropriate to do the merge. Also, if you're curious about the button layout for A/B, you'll find it here...

The A button is on the far right. B is the one at the bottom.

landfillbaby:
done :)

RobLoach:
Have you been able to test this on your Android device? Would like to make sure it's working before we merge it.

landfillbaby:
yes i tested and it's fine

PR #1206, opened 2024-12-08
Added Razer Raiju Tournament Edition Wired key mapping

Charging1948:
Tested and worked like a charm.


hizzlekizzle:
Looks great! While we have you, any chance we could get you to add the button labels, so it will display them accurately in the quick menu > controls instead of getting the generic retropad labels?

Here's an example of what that looks like: https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/Xbox%20360%20Wired%20Controller.cfg#L33

Charging1948:
I only tested wired mode for this controller earlier. Wireless mode is using different `vendor` and `product` ids. Should i also add a config for the wireless-mode?

Charging1948:
Wired one:
image

Wireless one:
image

Both working and with correct labels.

hizzlekizzle:
IIRC, RetroArch tries to match on device name first, and if it gets an exact match on that, it just goes with it, but if not, it falls back to vid/pid. 

If you want to change the name in the popup notification without affecting the matching, you can add the optional `input_device_display_name` field.

Charging1948:
Ok thank you for clarifying. I reverted the `input_name`

PR #1171, opened 2024-08-06
Update Sony_DualShock_4_Controller_v2.cfg

davidhedlund:
input_menu_toggle_btn = "110"
input_menu_toggle_label = "PS"

PR #1215, opened 2025-01-16
android: Update Sony DualShock 4 v2

6lackmag3:
This fixes current DS4 v2 android config with proper axis and buttons codes. Also fixes Quick Menu hotkey wrongly mapped to Select/Share button.


6lackmag3:
My apologies as I didn't see #1171

PR #1224, opened 2025-03-08
dinput: Add 8BitDo Ultimate 2C Wireless Controller BT

JonSnow88:
8BitDo Ultimate 2C Wireless Controller Bluetooth profile, so you can use without the usb dongle.


jackinloadup:
Thanks, I used this today! Worked with some basic testing playing MrBoom. confirm dpad and action buttons work.

RobLoach:
Thanks a lot! Would either of you be able to check the configs for SDL2 or any of the other input devices too? :pray:

PR #1227, opened 2025-04-05
udev: Update Nintendo Switch controllers after button remaps

nfp0:
Something somewhere changed the button mappings on these Nintendo Switch controllers, making the autoconfigs put the buttons on nonsensical positions.
I have checked and there has not been any recent change to the `hid-nintendo` kernel driver, so I don't know what caused this to change or when.

This PR gets them back to the logical positions they had before. But I'm marking this as draft because I would like to have at least one other person confirm that this is not something weird on my distro (I'm on Arch) and has indeed changed for everyone.

Additionally, I do not have the 6 button version of the Mega Drive/Genesis controller, so I can't confirm where the XYZ buttons will land, so I just mapped them to the next buttons in numeric sequence.
@PitiBatou I know you have the 6 button controller, so it would be super nice if you could confirm my mappings to be according to this:
https://retropie.org.uk/docs/Mega-Drive-Genesis/


RobLoach:
If we make the change, people may complain quickly, or not... Perhaps we just bring it in? I don't have a switch controller to test.

nfp0:
I have also tested on my other PC and it also has the same button mismatch there, so I guess this is universal and we can move forward with it, if no one else is able to test it.

PR #1229, opened 2025-04-17
Capcom Home Arcade Controller: Fix vendor_id CHA1

cjom:
Surprisingly, CHA1 and CHA2 don't have the same vendor_id.
CHA2 is correct, only CHA1 needs rectification.


cjom:
CHA1

PR #1246, opened 2025-07-18
xinput: Add 8BitDo Arcade Stick USB

HardDie:
Added profile for 8BitDo Arcade Stick connected via cable

https://www.8bitdo.com/arcade-stick/


sonninnos:
This conflicts with wired X360 with the same vid/pid, please fix or remove.

hizzlekizzle:
if it *only* has compatibility modes, I don't think there's any way we can provide an autoconfiguration file for it without breaking the actual controllers it's masquerading as.

PR #1251, opened 2025-07-25
Delete xinput/8bitdo_arcade_stick.cfg

### LINKED ISSUE #1249: 8bitdo_arcade_stick.cfg conflicts with ROG Ally Controls
It appears the 8bitdo_arcade_stick.cfg now takes over for the ROG Ally's Built In Controller.  I believe this is because the ROG Ally's built in controller is detected as "Controller (XBOX 360 For Windows)" like I assume the 8bitdo arcade stick is detected as.  Unfortunately the Input Mapping is completely different thus the ROG Ally would need to manually configure and not have use of Auto Config.

If this is helpful I can report the ROG Ally Detect VID and PID in Windows it seems to have 2 sets:
USB\VID_045E&PID_028E&REV_0114
USB\VID_045E&PID_028E

Please let me know if anything further questions.

hunterk:
We should not accept any 8bitdo configs for compatibility modes, only dinput/android (i.e., where it reports the controller's "real" vid/pid).

Closes #1249


HardDie:
Is there any way to fix this?
Oh maybe there is a way to find "real" vid/pid?

HardDie:
This arcade stick have two mode: xinput, nintendo switch.
In xinput values: VID 1118 (45E), PID 654 (28E)
In nintendo switch values: VID 1406 (57E), PID 8201 (2009)

I can't find any other VID/PID values for this controller.
Is there any other options how to check if it's X360 controller or Aracade Stick?

XerTheSquirrel (approved):

hizzlekizzle:
@HardDie see if it has a dinput/android compatibility mode. That's where 8bitdo stuff will typically report its actual vid/pid and name. The xinput and switch modes are compatibility modes where it lies about those things and just reports generic ones to make the console think it's one of their own controllers.

PR #1261, opened 2025-08-16
Add support for 8BitDo Ultimate C Wired Controller for Xbox (udev and sdl2)

Gragt:
The database already contains config files for the 8BitDo Ultimate C Wired Controller, but this a different, albeit similar, device.

`input_device`, `input_vendor_id`, and `input_product_id` were taken from `lsusb` output.


Gragt:
I just noticed another pull request from 2 weeks ago aimed to do the same thing (udev only, not sdl2).

Gragt:
I apparently can’t read. The other pull request was for another pad — notice the lack of the friendly capital C. The two config files proposed here should be okay to add. Sorry for the confusion.

PR #1262, opened 2025-08-17
Add 8BitDo Ultimate C Wired Controller for Xbox (sdl2)

Gragt:
This builds on [this commit](https://github.com/libretro/retroarch-joypad-autoconfig/pull/1254) for udev by providing sdl2 configuration.

`input_device`, `input_vendor_id`, and `input_product_id` are all taken from `lsusb` output.


Gragt:
Apparently, I can’t read. I thought the other commit was for 8BitDo Ultimate C Wired, but it is instead for 8BitDo Ultimate Wired. I’m a silly fellow.

PR #1270, opened 2025-10-01
udev: add Manba One BFM

YodaDaCoda:
This device has multiple connection modes. This is for Bluetooth connection while in Nintendo Switch mode. Haven't attempted to map the other modes yet.


hizzlekizzle:
Unfortunately, that vid/pid appears to have a lot of conflicts, so we would need to comment them out at least for now and operate purely on the device name.

Also, while we have you, can you add the button labels to the file, since you have the physical gamepad in your possession?

YodaDaCoda:
Edited OP - this is when connected in Android mode over bluetooth. Device doesn't seem to be recognised by retroarch in Switch mode (happy to take suggestions as to what might be going on there). Is recognised as xbox360 controller in PC mode and works as expected.

As requested, I've commented out the vid/pid and added the button labels.

Issue #814, opened 2022-01-04
Missing VIDs and PIDs


Jaffacakelover:
I've been through all of the controller autoconfig files for a related task, and thought this list might prove useful to some who want to track down missing values.

**The following autoconfigs have no vendor_id and no product_id:**

> **android:**
* ADC_Joystick
* Amazon Fire TV Remote
* Android TV Remote Control
* Archos_Gamepad
* BBC_GAME
* Buffalo_BGC_FC801
* ChromeOS_Gamepad
* Defender_Game_Racer_Classic
* DragonRise
* Drone
* DualShock2_Generic
* DualShock2_JCPS102
* DualShock2_WiseGroup
* DualShock3
* DualShock3_Gamepad0
* DualShock3_Gamepad1
* DualShock3_Gamepad2
* DualShock3_Gamepad3
* DualShock3_Gamepad4
* DualShock3_Gamepad5
* Elecom_JC-U912F
* GameMID
* GameMID_Alt
* Gasia_PlayStation3
* Genius_Maxfire_G08XU
* GPD_G5A
* GPD_G58_skeltonrom
* GPD_Q9
* GS_controller
* GS_gamepad
* HuiJia
* iControlPad_HID_Joystick_Profile
* iControlPad_SPP_Profile_Bluez
* idroid_con
* iDroid_x360
* Infra-Red
* JOYROOM-ZS149
* JXD_S5110B
* JXD_S5110B_Skelrom
* JXD_S7300B
* JXD_S7800B
* JXD_S7800B_1
* Logicool_RumblePad_2
* Logitech_Precision
* Logitech_RumblePad_2
* MadCatz_PC_USB_Wired_Stick
* Mayflash_Wii_Classic
* Microsoft_Sidewinder_Classic
* Microsoft_Sidewinder_Dual_Strike
* Moga_2_HID
* Moga_IME
* MUCH_iReadyGo_i5
* nvec_keyboard
* NVIDIA_SHIELD_Remote
* NVIDIA_SHIELD_Virtual
* Nyko_Playpad_Pro
* Onlive_Wireless_Controller
* OUYA
* PLAYSTATION(R)3 Controller
* RetroKeyboard
* retronicdesign.com Sega Joypad Adapter v3.0
* RetroUSB_RetroPad
* RetroUSB_SNES_RetroPort
* Saitek_RumblePad_480
* Samsung_Gamepad_EI-GP20
* Sega_Virtua_Stick
[...]

PR #1274, opened 2025-11-04
udev: Improve Wii Remote Support for RVL-CNT-01

fekir:
The remote `RVL-CNT-01` seems to work out of the box, but `RVL-CNT-01-TR` remotes have all buttons mapped incorrectly.

This issue can be avoided by providing the `vendor_id` and `product_id`

### DIFF
--- a/udev/Nintendo_Wii_Remote_TR.cfg
+++ b/udev/Nintendo_Wii_Remote_TR.cfg
@@ -1,9 +1,9 @@
-# Nintendo RVL-CNT-01
+# Nintendo RVL-CNT-01-TR
 input_driver = "udev"
-input_device = "Nintendo Wii Remote"
-input_device_display_name = "Nintendo Wii Remote"
+input_device = "Nintendo Wii Remote TR"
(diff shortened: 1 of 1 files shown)


fekir:
Also related to https://github.com/libretro/retroarch-joypad-autoconfig/issues/814

cscd98:
Clean install of RetroArch configs (nothing in ~/.config/retroarch) and using official Wii mote.

Launch RetroArch
Online updater - install controller updates etc.
Load latest dolphin core
Mario Kart Wii -> B and A are the wrong way around

At least that is my experience on Linux. You can't turn yet as I haven't looked into adding gyro

fekir:
You left out which remote you use in the the dolphin core, there are many possibilities: wiimote horizontal, passtrhough, gamecube controller, wiimote+nunchuck, ...

cscd98:
That option no longer appears for some strange reason. More debugging is required..

fekir:
> That option no longer appears for some strange reason. More debugging is required..

Strange indeed, but it does not seem to depend on this changes then, as they define only the mapping between the wiimote and retropad

PR #1284, opened 2025-12-29
dinput, xinput: Add Brawler64 + Micro

Jaffacakelover:
Added Retro Fighters Brawler64 and 8bitdo Micro.


sonninnos:
This needs to be reverted since it again conflicts with a real X360 wired controller: https://github.com/libretro/RetroArch/issues/18739

PR #1285, opened 2025-12-30
xinput, dinput, sdl: Add Flydigi Vader 5 Pro

butterbutts:
Adds support for the Flydigi Vader 5 Pro over both bluetooth and wireless usb dongle.


RobLoach (approved):
Looks okay to me.

PR #1311, opened 2026-04-09
android: Add 8BitDo Pro 2 (XInput Bluetooth)

TideGear:
The 8BitDo Pro 2 in XInput BT mode (mode switch set to X) spoofs
Microsoft VID 0x045E / PID 0x02E0, identical to the Xbox One Wireless
Controller. RetroArch matches the existing Xbox One profile on VID+PID
alone (score 60), since the BT advertised name "8BitDo Pro 2" differs
from the Xbox One profile's input_device "Xbox Wireless Controller".

This causes incorrect button mapping throughout: the Xbox One profile
expects Microsoft's proprietary BT protocol keycodes, which differ from
the standard HID-over-BT keycodes 8BitDo actually sends for nearly every
button except A and B. Verified via pad tester.

This profile matches on device name in addition to VID+PID for a total
score of 90, taking priority over the Xbox One profile, and uses the
correct standard HID keycodes throughout.

### DIFF
--- a/android/8BitDo_Pro2_XInput_BT.cfg
+++ b/android/8BitDo_Pro2_XInput_BT.cfg
@@ -2,17 +2,17 @@
 # Set the mode switch on the back of the controller to X (XInput).
 #
 # In XInput BT mode the controller spoofs Microsoft VID 0x045E / PID 0x02E0
-# (identical to the Xbox One Wireless Controller).  RetroArch matches the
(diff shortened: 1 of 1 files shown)


TideGear:
Also worth mentioning: controller vibration is now supported with BT XInput controllers in Android RA, DInput 8BitDo doesn't seem to be able to do vibration, necessitating the above.

RobLoach (approved):

PR #1312, opened 2026-04-10
sdl: dinput: Add 8BitDo 64 Bluetooth Controller

odditude42:
Adds dinput and sdl2 autoconfiguration profiles for the 8BitDo 64 Bluetooth Controller, connected either by USB or BT (the controller presents itself identically both ways).

Caveat: The controller presents the C-buttons as a pair of analog axes; the Up-C/Down-C and Left-C/Right-C buttons cannot be read simultaneously.

Tested on Windows 11 25H2 (b26200.8037) with RA input driver set to xinput/dinput/sdl2 (xinput uses the dinput profile).


odditude42:
Tested the controller in Switch mode:

* xinput driver: Inputs are constantly spammed, not debugging.
* sdl2 driver: C-buttons are not recognized at all.

Tested the controller with RobLoach's SDL3 build https://github.com/libretro/RetroArch/pull/18833 -

* sdl3 driver:
  * Controller in Dinput mode: C-buttons are recognized as independent buttons, but still function as opposing axes (e.g. Up-C and Down-C negate each other's input)
  * Controller in Switch mode: All controller buttons are recognized and function properly.

RobLoach (approved):

Issue #1277, opened 2025-11-12
Travis CI through GitHub services integration is deprecated


davidhedlund:
# Issue
Some PRs get stuck at a mandatory "continuous-integration/travis-ci" check

<img width="939" height="291" alt="Image" src="https://github.com/user-attachments/assets/ea8b6e10-a6f4-4803-8d4f-cce8731edffb" />

> 1 pending check
> continuous-integration/travis-ci Expected — Waiting for status to be reported

Example of an affected PR from a fresh fork that cloned today:
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1276

I encountered the same issue earlier this year when I used an older fork﻿. After a few months, I gave up and created a fresh fork and PR﻿ (after I discarded the commits from my previous PRs), as advised by hizzlekizzle﻿, hoping the problem would be resolved. However, the issue has just reappeared. I have not touched my repository settings at all in order to prevent further confusion.

The remaining opened PRs are also affected:
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1195
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1256
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1263
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1239
> Closing this merge request in favor of https://github.com/libretro/retroarch-joypad-autoconfig/pull/1255 since this one has weird conflicts on files this MR doesn't even touch that I don't want to spend any more time resolving.

This should be fixed in the repo﻿ because it is preventing contributions.

# Workaround
I emailed support@travis-ci.com and they replied within a few hours:

> Thank you for reaching out to Travis CI, and sorry for the hassle.
> 
> It seems like your GitHub Org and repositories are linked to Travis CI through GitHub services integration ("continuous-integration/travis-ci"), which is deprecated. Please install the Travis CI GitHub app (https://github.com/apps/travis-ci) for the Org account. Sync the account afterward and trigger a build to see if the status checks issue is resolved.
> 
> We will follow up on your reply.

An admin must
* Edit branch protection settings under [Settings Branches](https://github.com/libretro/retroarch-joypad-autoconfig/settings/branches)
  - Disable "continuous-integration/travis-ci"
  - Enable "Travis CI - Pull Request" or "Travis CI GitHub" if Travis is still required.
* Remove the old Travis check Comment on this PR.

Can you please do this? @LibretroAdmin @hizzlekizzle

PR #1318, opened 2026-06-16
android: Update L2/R2 as Buttons

agmun4o-pixel:
Correction of wrong buttons and labels


hizzlekizzle:
By changing them to 'btn' instead of axis, they can no longer provide analog control, right?

davidhedlund:
> By changing them to 'btn' instead of axis, they can no longer provide analog control, right?

That is correct.

Long before this PR, seven months ago, I spent days working on another PR for this controller mapping—which remains unmerged. Part of that work included thoroughly testing and finalizing the L2 and R2 **axis** values for accuracy.
* https://github.com/libretro/retroarch-joypad-autoconfig/pull/1276

@agmun4o-pixel As someone who has maintained the DualShock 4 v2 configuration for a while, I can assure you this is my final PR for the mappings. The delay in getting it merged seems to be tied to a broader open issue here:
* https://github.com/libretro/retroarch-joypad-autoconfig/issues/1277

PR #1322, opened 2026-07-13
sdl3: Update and add license

RobLoach:
The SDL3 Game Controller DB is zlib, so I believe we need to include the LICENSE? Stick it at the top? Or part of COPYING. Is there a better way to do this?


hizzlekizzle:
I think adding it to COPYING seems good. I'll mention it in that issue where a guy is asking about using our profiles in a commercial application: https://github.com/libretro/retroarch-joypad-autoconfig/pull/1322

PR #1325, opened 2026-07-23
android: Add Sunxi IR Remote

danielcamposramos:
## Summary

Add an Android menu-navigation profile for the Allwinner built-in IR receiver
that reports:

```text
Device: sunxi-ir-uinput
VID:    0x0000 / 0
PID:    0x0000 / 0
Source: DPAD | KEYBOARD (0x00000301)
```

The repository already has `sunxi-ir.cfg`, but Android treats
`sunxi-ir-uinput` as a different exact input device, so that profile does not
match this receiver.

## Hardware validation

- Android 10 TV box using an Allwinner H616 SoC.
- Physical bundled IR remote.
- Android's active `/system/usr/keylayout/sunxi-ir-uinput.kl` maps Back,
  D-pad directions, D-pad Center, and Menu to keycodes 4, 19–23, and 82.
- The physical remote was confirmed navigating RetroArch with arrows, OK,
  Back, and Menu.
- The profile deliberately includes only those validated menu functions.
- Repository test result: `No duplicate profiles were found`.

HDMI-CEC is intentionally excluded: on this Android build CEC events are
framework-injected as `deviceId=-1`, `SOURCE_HDMI`, with no physical device
name or VID/PID, so they cannot be represented honestly by a joypad
autoconfiguration profile.

### DIFF
--- a/android/sunxi-ir-uinput.cfg
+++ b/android/sunxi-ir-uinput.cfg
@@ -1,0 +1,24 @@
+# Allwinner built-in infrared receiver, intended for menu navigation.
+
+input_driver = "android"
+input_device = "sunxi-ir-uinput"
+input_device_display_name = "Allwinner TV Remote (IR)"
+input_device_type = "remote"
+input_vendor_id = "0"
+input_product_id = "0"
+
+input_b_btn = "4"
+input_up_btn = "19"
+input_down_btn = "20"
+input_left_btn = "21"
+input_right_btn = "22"
+input_a_btn = "23"
+input_menu_toggle_btn = "82"
+
+input_b_btn_label = "Back"
+input_up_btn_label = "Up"
+input_down_btn_label = "Down"
+input_left_btn_label = "Left"
+input_right_btn_label = "Right"
+input_a_btn_label = "OK"
+input_menu_toggle_btn_label = "Menu"


android/sunxi-ir-uinput.cfg
```diff
@@ -0,0 +1,24 @@
+# Allwinner built-in infrared receiver, intended for menu navigation.
+
+input_driver = "android"
+input_device = "sunxi-ir-uinput"
+input_device_display_name = "Allwinner TV Remote (IR)"
+input_device_type = "remote"
+input_vendor_id = "0"
+input_product_id = "0"
```
RobLoach, line 8: A vendor/Product ID of 0 confuses me. I understand that's what's reported, but does that just mean it errored out when getting the values?

danielcamposramos: Good question. No—the zeroes are not the result of a failed lookup.

**[proven]** Across three independent device captures, Android's Input Manager reported the complete identity tuple for this exact device as:

```
bus=0x0019, vendor=0x0000, product=0x0000, version=0x0004
```

Other input devices in those same captures reported nonzero vendor/product IDs, so the identity-retrieval path was working. The kernel registers `sunxi-ir-uinput` under `/devices/virtual/input/...`; the factory `multi_ir` binary also references both `/dev/uinput` and the `sunxi-ir-uinput` name.

**[inferred]** The userspace uinput producer leaves the vendor and product members of `input_id` unset, so Linux exposes their legitimate value as zero. I included `0` in the profile because it is the literal value Android reports—not because our capture errored.

If zero has unwanted wildcard semantics in the autoconfig matcher, I am happy to omit those two fields and rely on the exact device name instead. I can also add a short comment to the cfg clarifying why the IDs are zero.

PR #1329, opened 2026-08-15
udev: Add 8BitDo SN30 Modkit (BT)

TooTallNate:
The udev folder covers the M30/N30/P30/S30 mod kits but not the SNES/SFC one.

Button numbering matches the other D-input mode 8BitDo pads (vid:pid 2DC8:5103). The D-pad is mapped as both axes and hat0 since firmware revisions report it differently. Tested over Bluetooth on a Raspberry Pi 4 (kernel 6.6, hid-generic) with RetroArch master - all buttons and D-pad verified in-game.


hizzlekizzle:
looks good to me

PR #1332, opened 2026-08-24
8bitdo SN30 pro+ bluetooth configuration

bbbradsmith:
xinput autoconfig for an 8BitDo SN30 pro+ using bluetooth


hizzlekizzle:
Unfortunately, that's using a generic xbox vid/pid, so it'll false-match on all sorts of stuff. If you comment them out, though, we can merge this to match on name only.

bbbradsmith:
Okay, done.

hizzlekizzle:
lgtm, thanks!

PR #1333, opened 2026-08-30
udev: restore modern device names to Switch Pro Controller (non-HID)

rib3ye:
### The problem

This profile can't be turned on, because it looks for a controller name that current kernels don't use.

The file is default-off, so you enable it by uncommenting `input_device`, `input_vendor_id` and `input_product_id`, as the [docs](https://docs.libretro.com/guides/controller-autoconfiguration/) describe. But the only name it offers is `Nintendo Switch Pro Controller`, and modern kernels report a Bluetooth Pro Controller as just `Pro Controller`. So you follow the instructions and still get nothing.

### Why the names differ

The kernel changed what it calls these controllers:

- Roughly 5.15 to 6.2: `Nintendo Switch Pro Controller`
- 6.8 and newer: `Pro Controller` over Bluetooth, `Nintendo Co., Ltd. Pro Controller` over USB

The main `Nintendo Switch Pro Controller.cfg` lists all three names, so it still matches. This default-off file listed `Pro Controller` too, until e48f37a replaced it with the long name when the file was renamed. The two have been out of step since.

### Who this affects

Anyone whose Pro Controller isn't being driven by `hid_nintendo` — a kernel built without `CONFIG_HID_NINTENDO`, the driver blacklisted, or the driver failing to bind. That's what the non-HID profile is for, and genuine Nintendo controllers land there too. It isn't specific to any one brand of pad.

### The fix

Add the two newer names as commented alternates, so this file lists the same names the main profile does.

They stay commented out. Nothing changes for anyone unless they deliberately enable this profile, and no button mappings are touched.

### What I see here

Debian 13, Linux 6.12.105, connected over Bluetooth:

```
I: Bus=0005 Vendor=057e Product=2009 Version=0001
N: Name="Pro Controller"
```

So: name `Pro Controller`, vendor `1406`, product `8201`.

### Testing

RetroArch 1.20.0, Debian package, `udev` driver.

I took this file's existing button mapping, added `input_device = "Pro Controller"` with the same vendor and product IDs, and RetroArch matched it and set up both controllers:

```
[Autoconf]: ... configured in port 1.
[Autoconf]: ... configured in port 2.
```

### One thing worth flagging

My test controller is an 8BitDo SN30 Pro in Switch mode, not a genuine Nintendo pad. It reports the same IDs, and it runs under `hid-generic` on my machine because `hid_nintendo`'s startup handshake times out against that firmware.

The naming above comes from how the kernel names HID devices rather than from this particular controller, and the main profile already lists the same three names. But you should know what I tested on.


hizzlekizzle:
this seems fine to me.

PR #1334, opened 2026-08-31
Add 8BitDo Micro controller profile

tensaisuper:
Adds a udev autoconfig profile for the 8BitDo Micro
in Android (D-Input) mode.

The D-pad is exposed as HAT0X/HAT0Y axes.


hizzlekizzle:
This looks good to me.

Edit: Except for 'input_mode_btn". I believe that should be menu_toggle_btn, right?

tensaisuper:
You're right. I'll change it to input_menu_toggle_btn. Thanks for pointing that out.

